PR #41, opened 2015-01-25
Fixes to allow Compilation using VC++ 12

bodrick:
Some minor fixes and tweaks to allow compilation under Visual C++ 12.
This doesn't include changes that I need to make to the build
generation.


cjcliffe:
Looks good, thanks -- merged now.

PR #42, opened 2015-01-25
Updated cmake for Visual C++12

bodrick:
Modified cmake so that it properly generates for building under Visual C++ 12.  I also added the ability to build either 32bit or 64 bit builds, and added flag for windows builds that allow building either using WASAPI or DirectSound.


cjcliffe:
This looks great, going to give it a try and test tdm-gcc with 32 / 64 bit too.

cjcliffe:
Good news -- Using this I now have a CMake / lib configuration that builds with TDM-GCC and Visual Studio with both producing working builds.

Will be tweaking the config and pushing results shortly for some platform testing, and if all is good will merge to master.

Thanks again!

cjcliffe:
This is now merged to master -- I haven't had a chance to try out 32-bit yet but that's the next step.  Thanks for your help!

PR #49, opened 2015-02-03
Added LUT for IQ mapping with proper scaling

mryndzionek:
I've changed the IQ scaling to table lookup. Also updated the scaling:

See:
http://cgit.osmocom.org/gr-osmosdr/commit/?id=0edfcfcba0ddf0fd5002dab70331d75a376bcf3f

### DIFF
--- a/src/sdr/SDRPostThread.cpp
+++ b/src/sdr/SDRPostThread.cpp
@@ -2,11 +2,27 @@
 #include "CubicSDRDefs.h"
 #include "CubicSDR.h"
 
+#include <complex.h>
+#include <endian.h>
 #include <vector>
 #include <deque>
 
 SDRPostThread::SDRPostThread() :
         iqDataOutQueue(NULL), iqDataInQueue(NULL), iqVisualQueue(NULL), terminated(false), dcFilter(NULL), num_vis_samples(16384) {
+
+    // create a lookup table
+    for (unsigned int i = 0; i <= 0xffff; i++) {
+        liquid_float_complex tmp;
+# if (__BYTE_ORDER == __LITTLE_ENDIAN)
+        tmp.real = (float(i & 0xff) - 127.4f) * (1.0f/128.0f);
+        tmp.imag = (float(i >> 8) - 127.4f) * (1.0f/128.0f);
+        _lut.push_back(tmp);
+#else // BIG_ENDIAN
+        tmp.real = (float(i >> 8) - 127.4f) * (1.0f/128.0f);
+        tmp.imag = (float(i & 0xff) - 127.4f) * (1.0f/128.0f);
+        _lut.push_back(tmp);
+#endif
+    }
 }
 
 SDRPostThread::~SDRPostThread() {
@@ -82,8 +98,7 @@ void SDRPostThread::threadMain() {
             }
 
             for (int i = 0, iMax = dataSize; i < iMax; i++) {
-                fpData[i].real = (float) data_in->data[i * 2] / 127.0;
-                fpData[i].imag = (float) data_in->data[i * 2 + 1] / 127.0;
+                fpData[i] = _lut[*((uint16_t*)&data_in->data[2*i])];
             }
 
             iirfilt_crcf_execute_block(dcFilter, &fpData[0], dataSize, &dataOut[0]);
--- a/src/sdr/SDRPostThread.h
+++ b/src/sdr/SDRPostThread.h
@@ -32,4 +32,7 @@ protected:
     std::atomic<bool> terminated;
     iirfilt_crcf dcFilter;
     int num_vis_samples;
+
+private:
+    std::vector<liquid_float_complex> _lut;
 };
--- a/src/sdr/SDRThread.cpp
+++ b/src/sdr/SDRThread.cpp
@@ -256,7 +256,7 @@ void SDRThread::threadMain() {
         }
 
         for (int i = 0; i < n_read; i++) {
-            dataOut->data[i] = buf[i] - 127;
+            dataOut->data[i] = buf[i];
         }
 
         double time_slice = (double) n_read / (double) sampleRate;
--- a/src/sdr/SDRThread.h
+++ b/src/sdr/SDRThread.h
@@ -99,7 +99,7 @@ class SDRThreadIQData: public ReferenceCounter {
 public:
     long long frequency;
     long long sampleRate;
-    std::vector<signed char> data;
+    std::vector<unsigned char> data;
 
     SDRThreadIQData() :
             frequency(0), sampleRate(DEFAULT_SAMPLE_RATE) {


cjcliffe:
Thanks! -- this looks great and shows a decent performance improvement here; I hadn't done much with that code since the initial I/Q plotting tests.   Liquid's DC Blocking FIR filter is currently used to eliminate the DC offset but if this does the same I might be able to remove that stage.

I fixed a minor issue with OSX and merged to mryndzionek-master branch -- will test and finish merge on win32 when I get home tonight.

mryndzionek:
Awesome!
Great work altogether.

On my 3GHz I5 the app still consumes 100% CPU.
The profilers are showing that the hotspot is in libdri (`_mesa_make_temp_ubyte_image`).
Is this a Mesa problem?

Best regards,

cjcliffe:
Cheers -- thanks again.

I do find on my desktop in LInux I have to adjust the NVidia control panel settings before it works properly mainly by enabling then disabling vertical sync.  However the make_temp_ubyte_image sounds like the waterfall update -- do you get a decent improvement by changing the following in AppFrame.cpp?:

from: (line 57)
```
    demodSpectrumCanvas->setup(1024);
 ....
    demodWaterfallCanvas->setup(1024, 128);
```
 to
```
    demodSpectrumCanvas->setup(512);
 ....
    demodWaterfallCanvas->setup(512, 128);
```

and (line 102)
```
    spectrumCanvas->setup(2048);
 ....
    waterfallCanvas->setup(2048, 512);
```
to
```
    spectrumCanvas->setup(1024);
 ....
    waterfallCanvas->setup(1024, 512);
```

That will dramatically cut down the amount of texture memory being moved per-frame; eventually I may have it running in smaller static tiles for non-GPU or purely on the GPU in render buffers (so that call will be completely gone).  The problem right now is it  uploads 2048x512 + 1024x128 (around 1.6MB per frame) to the GPU, which will be compounded even more if it's running in Mesa software mode since the glPixelMap transfer function will multiply it by at least 3 and force the CPU to perform the waterfall colorization (which the GPU can do extremely fast).  Those tweaks above should reduce the texture bandwidth to ~590k per frame at the expense of some spectrum resolution beyond 1024 pixels wide view.

mryndzionek:
Changing the sizesto half of the values you suggested resulted in about 10% drop of CPU load (100%->90%).
No much improvement here...

cjcliffe:
@mryndzionek are you able to confirm it's using OpenGL hardware driver?  It seems that it might be using software rendering or possibly an unsupported feature/format which is causing software-fallback for some part of the rendering/texturing.  What's the GPU make/model and which driver are you using?

mryndzionek:
The `lspci` command outputs:

```sh
00:02.0 VGA compatible controller [0300]: Intel Corporation Xeon \
E3-1200 v3/4th Gen Core Processor \Integrated Graphics Controller [8086:0412] (rev 06)
	Subsystem: ASRock Incorporation Device [1849:0412]
	Kernel driver in use: i915
```

The `glxinfo` reports: `direct rendering: Yes`.
The `chrome://gpu/` section in browser reports that everything is hardware accelerated.
I've forced `cmake -DCMAKE_BUILD_TYPE=Release ..` and can see `-O3 -DNDEBUG` flags during build.

Using `INTEL_DEBUG=perf` I'm getting lots of:

```sh
Clear color unsupported by fast color clear.  Falling back to slow clear.
```

cjcliffe:
That GPU model should be fine -- the fast clear gives a decent boost for redraw speed but shouldn't be related to the image call.

I have just pushed an experimental waterfall-optimize branch (https://github.com/cjcliffe/CubicSDR/commit/679ad64a1f4dd7ec27e35f4a028723f36277c8e3) that removes all of the waterfall texture manipulation from the CPU.  The texture copy per frame should be roughly 2-4kb instead of 1.6mb which should almost completely eliminate heavy calls to _mesa_make_temp_ubyte_image.   Let me know if that helps!

mryndzionek:
Wow! I'm not actually an OpenGL guy and not able to exactly follow what did there but I'll let you know of the effect of this patch when I get home. Thanks for the opportunity to contribute to your project and learn something useful during the process! Creating something more self-contained and efficient is always a good idea.

mryndzionek:
Wow! Running with `vblank_mode=0` the waterfall speed is overwhelming and the CPU load jumps to 105% with the demodulator. Without vblank the load is at about 80%. The profiler shows that now the DC offset removal (`iirfilt_crcf_execute_block`) takes the most processing power. Removing this call:

```sh
             for (int i = 0, iMax = dataSize; i < iMax; i++) {
-                fpData[i] = _lut[*((uint16_t*)&data_in->data[2*i])];
+                dataOut[i] = _lut[*((uint16_t*)&data_in->data[2*i])];
             }
 
-            iirfilt_crcf_execute_block(dcFilter, &fpData[0], dataSize, &dataOut[0]);
+            //iirfilt_crcf_execute_block(dcFilter, &fpData[0], dataSize, &dataOut[0]);
```

results in 70% CPU load but upper waterfall shows some strange signals that are not visible on the main waterfall. The profiler shows that the demodulator thread consumes most of the CPU.
Most notably `dotprod_crcf_execute_mmx` called from `msresamp_crcf_execute` and `sincosf` called from `nco_crcf_mix_block_up`.
Well done altogether!

cjcliffe:
Nice, I'll give this a try after work and try to add in the frame-rate limiter.   Just the frame-rate limiter alone should tone down the dotprod and resampler calls, since they're the ones handling the zoom/fft that are now soaking up the previously unavailable CPU with no limits -- my Early 2011 3.2Ghz i7 + Nvidia GTX 560 is now rendering ~100fps in Windows 8 and over 300fps in linux -- that's a lot of wasted computation on imperceptible frames! :)

cjcliffe:
I've updated the waterfall-optimize branch, pull the latest changes and all platforms should now have a ~30fps (1/2 refresh rate in most cases = 60hz / 2) which is fixed for now until I add some options to adjust the rate.

I'm seeing approximately 14% -> 7% CPU usage drop in windows, and from ~40% overall cpu usage (at ~300fps) to ~15% overall cpu usage in Linux (Debian 7.8) now -- both when idle and no demodulation -- plus I haven't removed the DC FIR filter yet.

Let me know how that works for you, I wasn't sure which vertical sync GLX extension would be available on what systems so I implemeted the 3 that I could find and let it use the best/first available.

mryndzionek:
It does not work - there is no `GLX_SGI_swap_control`.  Here is the extension list:
```sh
Supported GL Extensions: GL_ARB_multisample GL_EXT_abgr GL_EXT_bgra GL_EXT_blend_color GL_EXT_blend_minmax GL_EXT_blend_subtract GL_EXT_copy_texture GL_EXT_polygon_offset GL_EXT_subtexture GL_EXT_texture_object GL_EXT_vertex_array GL_EXT_compiled_vertex_array GL_EXT_texture GL_EXT_texture3D GL_IBM_rasterpos_clip GL_ARB_point_parameters GL_EXT_draw_range_elements GL_EXT_packed_pixels GL_EXT_point_parameters GL_EXT_rescale_normal GL_EXT_separate_specular_color GL_EXT_texture_edge_clamp GL_SGIS_generate_mipmap GL_SGIS_texture_border_clamp GL_SGIS_texture_edge_clamp GL_SGIS_texture_lod GL_ARB_framebuffer_sRGB GL_ARB_multitexture GL_EXT_framebuffer_sRGB GL_IBM_multimode_draw_arrays GL_IBM_texture_mirrored_repeat GL_3DFX_texture_compression_FXT1 GL_ARB_texture_cube_map GL_ARB_texture_env_add GL_ARB_transpose_matrix GL_EXT_blend_func_separate GL_EXT_fog_coord GL_EXT_multi_draw_arrays GL_EXT_secondary_color GL_EXT_texture_env_add GL_EXT_texture_filter_anisotropic GL_EXT_texture_lod_bias GL_INGR_blend_func_separate GL_NV_blend_square GL_NV_light_max_exponent GL_NV_texgen_reflection GL_NV_texture_env_combine4 GL_S3_s3tc GL_SUN_multi_draw_arrays GL_ARB_texture_border_clamp GL_ARB_texture_compression GL_EXT_framebuffer_object GL_EXT_texture_compression_s3tc GL_EXT_texture_env_combine GL_EXT_texture_env_dot3 GL_MESA_window_pos GL_NV_packed_depth_stencil GL_NV_texture_rectangle GL_ARB_depth_texture GL_ARB_occlusion_query GL_ARB_shadow GL_ARB_texture_env_combine GL_ARB_texture_env_crossbar GL_ARB_texture_env_dot3 GL_ARB_texture_mirrored_repeat GL_ARB_window_pos GL_ATI_envmap_bumpmap GL_EXT_stencil_two_side GL_EXT_texture_cube_map GL_NV_depth_clamp GL_APPLE_packed_pixels GL_APPLE_vertex_array_object GL_ARB_draw_buffers GL_ARB_fragment_program GL_ARB_fragment_shader GL_ARB_shader_objects GL_ARB_vertex_program GL_ARB_vertex_shader GL_ATI_draw_buffers GL_ATI_texture_env_combine3 GL_ATI_texture_float GL_EXT_shadow_funcs GL_EXT_stencil_wrap GL_MESA_pack_invert GL_MESA_ycbcr_texture GL_NV_primitive_restart GL_ARB_depth_clamp GL_ARB_fragment_program_shadow GL_ARB_half_float_pixel GL_ARB_occlusion_query2 GL_ARB_point_sprite GL_ARB_shading_language_100 GL_ARB_sync GL_ARB_texture_non_power_of_two GL_ARB_vertex_buffer_object GL_ATI_blend_equation_separate GL_EXT_blend_equation_separate GL_OES_read_format GL_ARB_color_buffer_float GL_ARB_pixel_buffer_object GL_ARB_texture_compression_rgtc GL_ARB_texture_float GL_ARB_texture_rectangle GL_EXT_packed_float GL_EXT_pixel_buffer_object GL_EXT_texture_compression_dxt1 GL_EXT_texture_compression_rgtc GL_EXT_texture_rectangle GL_EXT_texture_sRGB GL_EXT_texture_shared_exponent GL_ARB_framebuffer_object GL_EXT_framebuffer_blit GL_EXT_framebuffer_multisample GL_EXT_packed_depth_stencil GL_APPLE_object_purgeable GL_ARB_vertex_array_object GL_ATI_separate_stencil GL_EXT_draw_buffers2 GL_EXT_draw_instanced GL_EXT_gpu_program_parameters GL_EXT_texture_array GL_EXT_texture_integer GL_EXT_texture_sRGB_decode GL_EXT_timer_query GL_OES_EGL_image GL_MESA_texture_array GL_ARB_copy_buffer GL_ARB_depth_buffer_float GL_ARB_draw_instanced GL_ARB_half_float_vertex GL_ARB_instanced_arrays GL_ARB_map_buffer_range GL_ARB_texture_rg GL_ARB_texture_swizzle GL_ARB_vertex_array_bgra GL_EXT_separate_shader_objects GL_EXT_texture_swizzle GL_EXT_vertex_array_bgra GL_NV_conditional_render GL_AMD_draw_buffers_blend GL_ARB_ES2_compatibility GL_ARB_blend_func_extended GL_ARB_debug_output GL_ARB_draw_buffers_blend GL_ARB_draw_elements_base_vertex GL_ARB_explicit_attrib_location GL_ARB_fragment_coord_conventions GL_ARB_provoking_vertex GL_ARB_sampler_objects GL_ARB_seamless_cube_map GL_ARB_shader_texture_lod GL_ARB_texture_cube_map_array GL_ARB_texture_multisample GL_ARB_texture_query_lod GL_ARB_texture_rgb10_a2ui GL_ARB_uniform_buffer_object GL_ARB_vertex_type_2_10_10_10_rev GL_EXT_provoking_vertex GL_EXT_texture_snorm GL_MESA_texture_signed_rgba GL_ARB_get_program_binary GL_ARB_robustness GL_ARB_shader_bit_encoding GL_ARB_timer_query GL_ANGLE_texture_compression_dxt3 GL_ANGLE_texture_compression_dxt5 GL_ARB_internalformat_query GL_ARB_shading_language_420pack GL_ARB_shading_language_packing GL_ARB_texture_storage GL_EXT_framebuffer_multisample_blit_scaled GL_EXT_transform_feedback GL_ARB_ES3_compatibility GL_ARB_invalidate_subdata GL_ARB_texture_storage_multisample 
```

The `libglx.so` has only these three swap-related functions: `DRI2HasSwapControl`, `DRI2SwapBuffers`, `DRI2SwapInterval`, `DRI2WaitSwap`. Is the `DRI2SwapInterval` suitable? Hadn't have time to look into this deeper.

cjcliffe:
It didn't show up as a standard GL Extension for me at all, but glxSwapIntervalEXT is the one that works on my NVidia card.  I'm not sure which cards even use the SGI extension it was just in the examples I found -- but DRI2SwapInterval sounds like the right function for you; it should be straightforward to add it to GLExt.cpp like the others -- just need to find the function definition for it then add dlsym call and elseif() check.

cjcliffe:
@mryndzionek removing the dcFilter code is causing the DC spike to return -- that's the signal you're seeing and it does badly interfere with tuning in that area as well as it seems to vary between my various models of dongles here.

That said, @jgaeddert mentioned he had a simpler method for removing the DC spike (from RTL-SDR devices in specific I believe) so I'll inquire about that as I'm sure it's probably more efficient than using the FIR filter.

cjcliffe:
@mryndzionek I've added a patch to waterfall-optimize that may work with DRI2SwapInterval.

jgaeddert:
It's pretty easy to remove a DC offset with liquid; create an IIR filter object with the "dc_blocker" create method:
iirfilt_crcf dcblock = iirfilt_crcf_create_dc_blocker(1e-4);

Then just pass every sample through this filter straight from the RTL SDR.

cjcliffe:
@jgaeddert thanks -- that's the one I've been using so far (and it works great) but something mentioned in prior conversation gave me the idea that you might have had a less expensive method that wasn't using the IIR filter.

mryndzionek:
@cjcliffe the `DRI2SwapInterval` is not an exported symbol in the `libglx.so` library. After searching for other candidate in my system I've found `eglSwapInterval(EGLDisplay dpy, int interval)`. It in the `libEGL.so` library.
However using it like so:
```sh
EGLDisplay dpy = glXGetCurrentDisplay();
 eglSwapInterval(dpy, interval);
```
does not seem to have any effect...

cjcliffe:
@mryndzionek did you dlopen the libEGL.so and use dlsym to obtain eglSwapInterval?  there were a couple functions that were exposed through GLX such as glxSwapIntervalMESA and glxSwapIntervalEXT that had no effect on my system; only the functions loaded through dlsym worked.   

Not sure why -ldl was required on your system and not mine but I'll add it to the appropriate cmake section.

jgaeddert:
Do you feel that the IIR filter is too expensive? Computationally it's much more efficient for this sort of thing than an FIR filter; an IIR filter may have an arbitrarily low bandwidth for a given filter length while a FIR filter cannot. To get the same effect with an FIR filter you'd need a lot more computation.
 
Alternatively it sounds like you're trying to characterize the DC component of the RTL device itself and then simply subtract it out as a static value. I've done this in the past, but these values might change based on frequency (hence your look-up table I assume), and even temperature.
 
I think I could probably write a calibration script that would simply sweep through a bunch of frequencies, determine the DC offset, and then store them in a file for later reference. Is this closer to what you're looking for?

cjcliffe:
@jgaeddert I haven't experienced it causing any real performance issues here; @mryndzionek had noted that iirfilt_crcf_execute_block() appears to now be at the top of the profile stack -- I think it really just means the rest of the code is running much better and no longer buried under the rendering overhead.

And now that you mention it, the DC component characterization with simple subtraction is what I believe you had referred to in our earliest conversations -- but if it requires calibration for each device/antenna/frequency/temperature  then it's probably not what I'm looking to do, since the IIR filter is already working perfectly and requires no configuration.

On a side(band) note I did some digging and found a major typo in my SSB implementations where I was actually using the wrong demodulation type altogether and had tweaked the input enough to make it function in some convoluted way -- I've gone back to give it another shot and learn some more; so you can feel free to disregard my last e-mail about it :-)

jgaeddert:
Just FYI, I wrote a quick script this morning to compute the DC calibration points swept over the entire range of my RTL-SDR. I'm attaching a plot that shows the real/imag DC offset points from 24 MHz to 1766 MHz with the antenna disconnected. For the most part the offset is constant (approximately -0.13 for both I and Q) but there are a few frequencies where the offset is much higher. Consequently I recommend keeping the DC blocking filter in place to help eliminate any offsets that can't be calibrated out with a simple look-up table.
rtlsdr_dc_calibration

cjcliffe:
@jgaeddert that's awesome thanks; I'm sure others would be interested too -- I'd throw that one up on the blog :+1:

mryndzionek:
@cjcliffe I've tried the dynamic loading of the EGL functions, but I it also does not work. I get EGL error 12296  - EGL_BAD_DISPLAY. I think we should continue further discussion [here](https://github.com/cjcliffe/CubicSDR/issues/52).
Best regards

jgaeddert:
Just for kicks I added a DC-blocking filter and swept through the band to see how it compares. The results are pretty good.

    iirfilt_crcf dcblocker = iirfilt_crcf_create_dc_blocker(1e-3f);
    {
        // inside loop:
        float complex x; // input sample
        float complex y; // output sample
        iirfilt_crcf_execute(dcblocker, x, &y);
    }
    iirfilt_crcf_destroy(dcblocker);

rtlsdr_dc_correction

PR #67, opened 2015-04-03
Frequency controls

cjcliffe:
- New tuning controls, spin-control style
- Hold shift while tuning to prevent carrying
- Follow demodulator while tuning
- Waterfall and spectrum view now update center frequency properly while zoomed
- Tuning bar help tips


mmgarland3:
How hard would it be to add mousewheel support for freq dials?

Also, ppm drift correction/offset is something CubicSDR is lacking, we need it for better precision before bookmarking is really useful (otherwise, bookmarks will not be precise, as the freq will chance as the chipset warms up)

cjcliffe:
I've added issue #68 for PPM correction, the first point can likely be done in short time.

PR #107, opened 2015-07-11
Advanced demodulators

Dantali0n:
This had advanced demodulators and a buffer for it just like the one for audio.
Maybe it would be better to pull this to a branch specific for the demodulator lab.

the demodulated data still needs to go trough some kind of mode decoder, I was thinking of creating a class similar to audiothread but then for digital data.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,10 @@
+CMakeCache.txt
+CMakeFiles/
+Makefile
+cmake_install.cmake
+compile_commands.json
+nbproject/
+x64/
 cmake_build/
 cmake_build_msvc/
 dist/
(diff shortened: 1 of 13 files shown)


cjcliffe:
This is looking good; I'm going to have to do some refactoring so that this can be broken up into more manageable and extendable pieces -- but for now if it's workable I'll give it a go.

I'll pull this into a new branch soon as there's several others that are nearly ready to merge including the ui_controls that will be the base for the demodulator lab, reusable visuals and a raw i/q demodulator mode that will likely need conflict resolution with this on a few lines.

Thanks! :+1:

cjcliffe:
This is now merged into an active branch called 'digital_lab' which I'll be working on to refactor the code and prototype the demodulator lab.

PR #155, opened 2015-10-08
Fixes for two crashes

eoineoineoin:
Hit a pair of crashes when using CubicSDR

* If rtlsdr_open fails, devTest is left uninitialized
* If the audio thread setActive() is called before run(), we dereference a null inputQueue.


cjcliffe:
Looks good; thanks -- I'll have to make sure I didn't make the same error in SoapyRTLSDR as well.

PR #181, opened 2015-11-09
fixed wxWidgets build issue

jocover:


cjcliffe:
Ah right, forgot to add this thanks :)

PR #188, opened 2015-11-15
Merge Digital Lab to Master

cjcliffe:
- Adds "ENABLE_DIGITAL_LAB" cmake bool and definition


Toontje:
Woohoo!! Now we are talking!! Building now. Can only start testing as of tomorrow afternoon. I will be on your back again.

cjcliffe:
@Toontje lol, it'll be a bit until we can do anything really useful with the output, but I'd be interested to see what digital locks you find and if any start dumping bits to the console.

Will add a wx output console soon.

vsboost:
Looking forward to playing with this as well.
Thanks

vsboost:
Dumb question but i just complied the latest and don't see anything different, is there anything i need to know?

Toontje:
@vsboost When you build add -DENABLE_DIGITAL_LAB=1 to your cmake statement and you will see this:
screen shot 2015-11-16 at 01 31 15
Didn't get it to do a lot though...

vsboost:
Ahh thanks for that, ill rebuild and see how we go.

Toontje:
I got this, but no data being dumped anywhere. I guess i'll wait (impatiently) for the next commit. ;-)
screen shot 2015-11-16 at 00 38 51

vsboost:
I got this.
image

i added the flag here in cmake for windows

image

vsboost:
Got it going on ubuntu, would still like to get it going on windows though, not sure whats going on during the build process.  I too am not seeing any data, tried an OOK transmission.

Dantali0n:
Data is not outputted anywhere yet, I had some tests with outputted it into the system console but don’t think there in their now. 
As for the lock / no lock that’s currently based on signal to noise ratio and can be inaccurate.

cjcliffe:
The primary reason for merging this is I'm half-done rewriting the Demodulator code so that every modem type is in it's own digital/analog 'modem' module -- once that's complete I think we'll be able to start doing some serious experiments with the digital code.

PR #212, opened 2015-12-05
Add ReBuffer buffer count warning, fix SDRPostThread off-by-one error

cjcliffe:
This wasn't the original problem, but once a demodulator hit the right-edge filterbank channel it would never de-activate and chew up the I/Q bandwidth memory refcounts for that channel if inactive.

### DIFF
--- a/src/sdr/SDRPostThread.cpp
+++ b/src/sdr/SDRPostThread.cpp
@@ -5,7 +5,7 @@
 #include <vector>
 #include <deque>
 
-SDRPostThread::SDRPostThread() : IOThread() {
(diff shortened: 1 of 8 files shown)


src/sdr/SDRPostThread.cpp
```diff
@@ -238,7 +238,7 @@ void SDRPostThread::run() {
                     firpfbch_crcf_analyzer_execute(channelizer, &data_in->data[i], &dataOut[i]);
                 }
 
-                for (int i = 0, iMax = numChannels; i < iMax; i++) {
+                for (int i = 0, iMax = numChannels+1; i < iMax; i++) {
```
cjcliffe: This was the actual bug :)

PR #241, opened 2016-01-01
Add peak hold to spectrum visuals

cjcliffe:
- Added peak hold support to SpectrumVisualProcessor
- Added peak hold toggle button
- Update SpectrumPanel to support peak hold display


vsboost:
Working well, thanks so much for this.

PR #265, opened 2016-01-27
Cleanup unused var warnings

viraptor:
Clean up compiling output so that actual issues can be spotted. Remove multiple
cases of unused parameter. Remove unused isActive at startup. (demod is started
regardless)


cjcliffe:
nice, thanks :+1: didn't know you could omit the var name to get rid of the unused param warnings :)

PR #266, opened 2016-01-27
Remove invalid const

viraptor:
const int return value doesn't really exist / is ignored. Fix a warning.


cjcliffe:
looks good, thanks!

PR #267, opened 2016-01-27
NetSDR related patch (better stream handling)

cjcliffe:
- Make readStream always request the stream MTU
- Deal with extra samples within CubicSDR


cjcliffe:
This seems to improve device performance overall here; I can't see a reason not to merge it in.

PR #268, opened 2016-01-28
Warning cleanups

viraptor:
All signed/unsigned bits were looked at separately and should be correct.
Majority of the project now compiles without warnings.


cjcliffe:
Looks great thanks! I'd had it on my mind to get this stuff cleaned up, but you're doing better than I'd planned :+1:

Issue #141, opened 2015-09-05
Keyboard shortcut enhancements


Toontje:
- [x] Centralize all keyboard input to AppFrame; Waterfall canvas is handling most of the work for the others at the moment
- [x] Show tooltips by default; allow users to disable them
- [x] Remove A/Z zoom (don't really need 4 ways to zoom :)
- [x] Add basic modem keys: A for AM,  F for FM, L for LSB, U for USB
- [x] Automatically start input dialog when typing a number (instead of pressing space first)
- [x] Allow manual entry of waterfall speed, gain, etc. via hovering and pressing space/typing

Toontje:
Found something i have not seen before. How many of those "easter eggs" am i still missing? ;-)

https://www.dropbox.com/s/9dvshlhlbjghljg/new.mov?dl=0

cjcliffe:
I've started working on these in the keyboard_updates branch; some of it working already.

cjcliffe:
@Toontje I tried doing the '?' showing a help popup but it was rather clunky and hard to reference to what you were currently doing.

During the process I had the idea to do something a bit more helpful for new users and make the hover tips from the status line show as a standard ToolTip popup beside the mouse when you hover the various parts of the application.  

I've also added first Settings menu option as "Enable Hover Tips" so that once you've learned what you need you can turn them off and by that time have probably noticed they're the same tips as in the status line anyways.  This setting will be saved.

Hopefully this helps new users to get going immediately to understand the features and experienced users will likely find it annoying after a few minutes and just turn it off.

Toontje:
I thik this actually works fine. And yes, this is for starters only. Once you know how to use the program you would want to switch it off. So good solution. Thanks!

PR #279, opened 2016-02-07
Keyboard Improvements

cjcliffe:
Related to Issue  #141
- [x] Create global keyboard handler
- [x] Remove A / Z zoom
- [x] Add A = AM, F = FM/FMS, U = USB, L = LSB
- [x] Start entry when a number is pressed
- [x] Show tooltips by default; allow users to disable them
- Add additional space-to-enter / numeric key:
  - [x] Waterfall Lines-Per-Second
  - [x] Spectrum Avg. Speed
  - [x] Gain Values
- [x] Resolve any platform issues and system key error noises

### DIFF
--- a/src/AppConfig.cpp
+++ b/src/AppConfig.cpp
@@ -198,6 +198,7 @@ AppConfig::AppConfig() : configName("") {
     winW.store(0);
     winH.store(0);
     winMax.store(false);
+    showTips.store(true);
     themeId.store(0);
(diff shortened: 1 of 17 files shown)


a-raccoon:
Allow numeric keypad (numpad) frequency inputs, and not just top-row numbers.  For the longest time I thought that keyboard frequency entry was broken (years) since I always use the numpad.

PR #356, opened 2016-06-01
Audio vis output queue nullptr crash and misc changes

vsonnier:
Hello @cjcliffe here is my first ever pull request in GitHub:)
There are several commits in this pull request:
-  FIX is the real essential fix for the crashes I had very often, due to a really unprotected concurrent modification of the audio vis output queue,
- MISCn are various suggestions of improvements, that you may consider or discard at will.

All commits are orthogonal, and the affected code has enough coments to be  self-explanatory I guess.


cjcliffe:
@vsonnier this looks great -- CubicSDR is only about my third attempt at threading something this complex so I'm sure there's probably still a lot of threading problems lurking :)

PR #358, opened 2016-06-02
Session and Crash fixes

cjcliffe:
- [x] Attempting to pinpoint session loading isssue
- [x] ReBuffer is overkill for scope vis and crashes OSX; just push and delete at reciever
- [x] Cleaned up lazy exception-fallback loadSession code
- [x] Clean up unnecessary stdout logging

### DIFF
--- a/src/demod/DemodulatorInstance.cpp
+++ b/src/demod/DemodulatorInstance.cpp
@@ -11,17 +11,19 @@ DemodulatorInstance::DemodulatorInstance() :
 #if ENABLE_DIGITAL_LAB
     activeOutput = nullptr;
 #endif
-	terminated.store(true);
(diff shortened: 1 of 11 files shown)


src/demod/DemodulatorInstance.cpp
```diff
@@ -11,17 +11,19 @@ DemodulatorInstance::DemodulatorInstance() :
...
+    terminated.store(true);
 	demodTerminated.store(true);
+    audioTerminated.store(true);
 	preDemodTerminated.store(true);
 	active.store(false);
 	squelch.store(false);
     muted.store(false);
-	tracking.store(false);
-	follow.store(false);
+    deltaLock.store(false);
```
cjcliffe: Oops..

PR #359, opened 2016-06-03
Replace mutex lock/unlock pairs with guards, cleanups

vsonnier:
@cjcliffe  Hi, another pedantic cleanup from me:
Locks / Unlock mutex pairs in code are quite noisy to use, and std::lock_guard can replace them almost everywhere:
-  Less code clutter
-  Less error prone
-  Exception safe
-  Garanteed unlocking order for nested guards
-  The real reason: raw mutexes are so 1990 !  

So I scanned the code for unlock() and replaced them with guards at the right places. (or so I hope)

I've also put additional guards at some places where the existing locking didn't seemed sufficient to protect the whole class consistency.

### DIFF
--- a/src/AppConfig.cpp
+++ b/src/AppConfig.cpp
@@ -47,39 +47,39 @@ bool DeviceConfig::getAGCMode() {
 
 
 void DeviceConfig::setDeviceId(std::string deviceId) {
-    busy_lock.lock();
+    std::lock_guard < std::mutex > lock(busy_lock);
     this->deviceId = deviceId;
-    busy_lock.unlock();
+    
 }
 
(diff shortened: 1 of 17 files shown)


vsonnier:
And sorry for the accidental reformating...

cjcliffe:
Looks great thanks; I did this with Demod Mgr and SDRPost Thread so I think it makes perfect sense to clean up the rest.

Going to give this a go on Mac and Linux and if all is good will merge it in.

cjcliffe:
@vsonnier everything looks stable here and no new leaks; going to merge it in

Thanks! 👍

PR #365, opened 2016-06-06
Update some external libs

vsonnier:
@cjcliffe This is a simple update from 2014-ish to 2016-ish versions of some external libs:
-  RtAudio 4.12,
-  TinyXML v2.62 (most recent from Sourceforge, dating from 2011), I had to rename some constants in DataTree.cpp but nothing dramatic,
-  LodePNG 20160501 from website.

Well at least for Win64 binaries it appears to work as well as before:)

### DIFF
--- a/external/lodepng/lodepng.cpp
+++ b/external/lodepng/lodepng.cpp
@@ -1,7 +1,7 @@
 /*
-LodePNG version 20141130
+LodePNG version 20160501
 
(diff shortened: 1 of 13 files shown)


cjcliffe:
Thanks @vsonnier -- All looks good at first glance; will pull here to test and merge shortly.

PR #367, opened 2016-06-08
Threading cleanup; fix some old mixer integration cruft and resolve a lot of random crashes..

cjcliffe:


src/audio/AudioThread.cpp
```diff
@@ -393,6 +391,20 @@ void AudioThread::run() {
...
+    // Drain any remaining inputs
+    if (inputQueue) while (!inputQueue->empty()) {
+        AudioThreadInput *ref;
+        inputQueue->pop(ref);
+        if (ref) {
+            ref->decRefCount();
+        }
+    }
+    
+    if (currentInput) {
```
cjcliffe: The actual bug here..

Issue #342, opened 2016-05-05
Feature request: Make Peak Hold display also for current demod view


vsonnier:
I don't know if that would be really usefull or not, since I'm no specialist: 
Make the Peak Hold option also display  a peak hold on the detailed, current demodulator view which looks more precise and reflecting more the "real" shape of the signal.

cjcliffe:
@vsonnier good call; it's already supported by that view I've just never thought to enable it.  Adding this to the enhancements, thanks!

vsonnier:
I've implemented it in https://github.com/cjcliffe/CubicSDR/pull/368. It just works.

cjcliffe:
looking good thanks!  I've also just added 'P' as the keyboard peak-hold toggle key.

vsonnier:
Indeed, P as global shortcut is convenient. Thanks !

PR #368, opened 2016-06-09
Peakhold on demod spectrum and misc

vsonnier:
@cjcliffe  Given your advice, I've done it in PEAK commit for issue https://github.com/cjcliffe/CubicSDR/issues/342.

And yes, MISCx commits are there also to salt this with completly irrelevant changes.


cjcliffe:
Looks good thanks; I figured that would work as-is but wasn't sure if the toggle should be separate -- this works fine though :)

Issue #301, opened 2016-02-25
Feature Request: Name Modems


george-viaud:
It would be useful to put a name on a 'modem' so it can be easily identified when within range.  Saved sessions would also record this name in the XML.

cjcliffe:
I'll close this up for now but i'll be left in reference as this will be covered by #161 -- naming them is inherently tied to bookmarking them.

Issue #161, opened 2015-10-14
Bookmarks and History log


Toontje:
Bookmarks / History / Active session log.
- [x] Create a dialog for bookmarks and history
  - [x] Use tree view
  - [x] Allow sorting by group
  - [x] Allow managing groups
- [x] Split active (live session) and bookmarks sections in tree
  - [x] Activate / bookmark in respective section to transfer from active to bookmark
- [x] Quick search all
- [x] Add "Recent" section for re-creating historical modems for that session
- [x] Allow bookmarking "Recent" and "Active"
- [x] Method to clear recent list
- [x] Method to remove anything from list

cjcliffe:
Feel free to curate the checklist for this issue; I think your original suggestion is in there somewhere :)

AlmightyOatmeal:
I would absolutely love the ability to save bookmarks and be able to group them.

cjcliffe:
@AlmightyOatmeal already partly working in the bookmarks branch (minus actually saving); hopefully will have some early tests out in the next few weeks.

cjcliffe:
@Toontje OSX test build is up at https://github.com/cjcliffe/CubicSDR/releases/tag/0.2.1-alpha1 
Other platform builds will be ready soon

Toontje:
Perfect implementation of this function. For me you can close this issue.

cjcliffe:
@Toontje thanks 👍 closing this up; we can make new ones for specific issues.

PR #369, opened 2016-06-11
Add an editable demod label visible on waterfall

vsonnier:
@cjcliffe  Hello again ! Since I woke up at 4am, why not doing some code ? Well here it is :

By now the demod type is displayed for each demod on the lower part of the waterfall.
This change enhance it with a user-configurable label, editable, saved and reloaded properly in the session file, using a new tag:  
````
<user_label>a useless description</user_label>
````
This allow to save a short description of a demod to quickly recognize it among several ones in a single session file. 

This actually looks like this:
cubicsdr

On demod focus, the 'E' key allow to edit the label.  The user label is displayed in such a way that the demod type is always on top. Then, clear the label with the Edit dialog and the user label desappears.

This enhancement is somewhat related to  https://github.com/cjcliffe/CubicSDR/issues/161 but this particular change wasn't listed there. On the other hand I found the issue https://github.com/cjcliffe/CubicSDR/issues/301 that describes the same thing I implemented.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -271,6 +271,7 @@ SET (cubicsdr_sources
 	src/AppFrame.cpp
 	src/AppConfig.cpp
 	src/FrequencyDialog.cpp
+    src/DemodLabelDialog.cpp
     src/IOThread.cpp
     src/ModemProperties.cpp
 	src/sdr/SDRDeviceInfo.cpp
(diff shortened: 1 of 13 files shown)


vsonnier:
The only trouble is that now that we can display any 8bit character, we can see that the existing GL font do not have all Latin1 characters. 
Using accented charaters like 'é' or 'ç' don't show up in display. Luckily GLFont skip them properly without crash.

cjcliffe:
This looks great; going to give it a try -- We could easily add the Latin1 character set as I simply used the BMFont generator for the font PNG images and it automatically maps to whatever characters you select for export in the BMFont file.

vsonnier:
Glad you like it ! :)

PR #370, opened 2016-06-12
Basic > 7bit strings support, now for user labels

vsonnier:
Hi @cjcliffe,  this is sort of continuation of https://github.com/cjcliffe/CubicSDR/pull/369 to be able to display > Ascii characters.

I wanted initially to add Unicode (well, std::wstring really) everywhere: I saw that it was too big, and mostly useless, because we only need std::wstring support where the user can see it and change it, for now only user labels.

At the end it was a half success, because DataTree do not understand > 8 (7?) bit characters...

So the modifications were:
-  GLFont full std::wstring support: font filenames, .fnt 16bit content, ability to render std:wstring...
-  PrimaryGLContext able to render user_labels with wstring characters,
-  DemodulatorInstance with std::wstring user labels,
-  Fonts were regenerated to include these 254 characters, using [Vera Font v1.10](http://ftp.gnome.org/pub/GNOME/sources/ttf-bitstream-vera/1.10/):

bmpfont

You can now toy with more esoteric labels at runtime, unfortunately they are converted back to Ascii (using '_' replacing > 7 bit charaters) on the session files in order not to break session loading completly.

Being a matter of pseudo-Unicode-ization, this may well break outside Win64, or create a black hole beneath your house, who knows ?


cjcliffe:
This looks great; I think we could do the 16-bit string type with xml support using HTML-type unicode conversion like```&#9824;``` or ```&#x2660;``` for 16-bit characters instead of attempting to store the actual characters; there may be some support in tinyxml already.

vsonnier:
Clever ! TinyXML supports UTF-8 already so I think the trouble is on DataElement which apparently only supports unsigned char*...so the translation should have to occur beetween User wstring and the DataTree model itself...

cjcliffe:
Been (successfully) busy getting AirSpy and UHD support working on windows; going to try and pull this in before the rc2 test build. 

I can add the element type to my DataTree class without much trouble; pulling this to have a look -- if I can patch it in easily now I'll give it a go.

cjcliffe:
looks good and merged in now with datatree support for wstring; thanks!

vsonnier:
Sorry for the additional work it has gave to you, and the portability issues. I warned it could open a black hole, though :)

cjcliffe:
Hah, no problem it seems like it needed to be done; thanks for your efforts :)

PR #371, opened 2016-06-13
wstring encode / decode fixes

vsonnier:
@cjcliffe Hi, I've noticed that the encode/decode/storage in DataTree was quite fucked up, with desappearing symbols and so on...

I think it should be OK now, even if we are system / codepage dependent with mbstowcs() / wcstombs(). 
For instance, on my Win7x64 French, these functions are not enable to convert the euro symbol: '€' even if all specific characters are now OK, at least as far as a French keyboard is concerned.

Well, we have opened the Pandora box of characters encoding, but for me I'll stop at that if you don't mind :)

### DIFF
--- a/src/AppFrame.cpp
+++ b/src/AppFrame.cpp
@@ -1562,8 +1562,6 @@ void AppFrame::saveSession(std::string fileName) {
         *demod->newChild("frequency") = (*instance_i)->getFrequency();
         *demod->newChild("type") = (*instance_i)->getDemodulatorType();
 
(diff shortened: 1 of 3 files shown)


vsonnier:
The Euro problem:
http://stackoverflow.com/questions/6368071/wcstombs-s-fails-when-input-string-is-%E2%82%AC-ascii-128d-in-c-cli

cjcliffe:
Sounds good; was just a quick attempt -- I guessed you'd find and fix some issues; thanks ;)

PR #372, opened 2016-06-14
Last string vs. wstring cleanups (hopefully)

vsonnier:
Well, I did a pass on wstring usage vs. string conversions, and found some inconsistencies, even if no bug had shown up.

However, this time I used wxString to do the proper handling, it was much easier than using mbstowcs() / wcstombs().


vsonnier:
I promise it is the last pull-request for this week :)

cjcliffe:
Hah, keep 'em rolling if they're improvements or fixes :)  Thanks again.

PR #376, opened 2016-06-19
FIX: non-Ascii and beyond are working better now (ex: Euro symbol)

vsonnier:
@cjcliffe  I hope this is the last fix of its kind: after reading much of Unicode lore on the Web, I tried the following one-liner, which actually fixed the Euro symbol problem, and likely others of more esoteric character sets.


cjcliffe:
Looks like a reasonable fix; nice to hear it gets around the issue 👍 

Thanks again!

Issue #352, opened 2016-05-27
font size


imesg:
Read a thread about font size which was closed Jan 13. I did not understand how to adjust the fonts but I did move the separators with no effect. I have a dell xps 13 with a 3800x1800 screen. This may not be a CubicSDR issue but the fonts are too small to read with the naked eye unless I get about 5 inches away from the screen. I need a magnifier to read the screen. I can read the frequencies below the 'Audio Visuals'. The menu bar is OK also. I struggle with this in other programs also. I am trying to sort the font issue out. If you have any suggestion for adjusting the fonts it may help me.

Thanks

cjcliffe:
Ah yes; I'll leave this open but I haven't had a chance to work with a retina screen; I can attempt to make some assumptions and simulate it here and see how it goes though.

imesg:
I guess I should have mentioned I have installed this on gentoo linux. CubicSDR really looks good on this display except for font size which may be a OS scaling issue.
Thanks

cjcliffe:
@imesg can I get a screenshot of the window maximized?  I think I can use that to approximate the next font size up that would be needed to make it appear properly at that resolution.

imesg:
cjcliffe did you get the screenshot I sent?

vsonnier:
I'm working on a way to use bigger menu settable fonts on a branch of my cloned repo. @cjcliffe I can work on it and you'll have a look when it is ready.

cjcliffe:
@vsonnier that would be great; even if there was just a simple small/med/large option. 👍

vsonnier:
@cjcliffe That would be normal (like now) and médium (1.33x) and large 2x for instance. l'm ok to take this issue, so that you can concentrate on more important matters like doing à SoapyAirspy module. 😃

ElectronicBattle:
hello

When I do ./CubicSDR.AppImage

I get a fault:

./CubicSDR.AppImage: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version 
`CXXABI_1.3.8' not found (required by ./CubicSDR.AppImage)
./CubicSDR.AppImage: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version 
`GLIBCXX_3.4.20' not found (required by ./CubicSDR.AppImage)
./CubicSDR.AppImage: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version 
[...]

PR #378, opened 2016-06-23
Settable font sizes with menu

vsonnier:
@cjcliffe Hello there. This is an attempt to help with issue #352, by adding ability to have different configurable font sizes, especially for 4K screens.
Note that I tested on a 1920x1080 screen.

Modifications:
-  Add a menu "Display" offering to tune the font size: Normal (1x) , Medium (1.5x), Large (2x) scale.
-  Font scale is applied on dynamic text (scope, floating labels, dBs...)

In addition, some graphical elements were made a little bigger, but are remained untouched by the zoom: (also because I didn't manage to find a way to do it properly !)  
-  Mode tiles get from 40pix to 44pix large, now displaying 18pix font instead of 16pix: quite clearer to the eye, and not wasting much more space,
-  TunerIndexedBar now displays a 32pix font instead of 24pix while fitting in the same space,
- All toggles 'V', 'S', 'M'...etc buttons are now 18pix instead 16pix, still fitting in the same space.

On the API side, GLFont::getFont no longer access fonts directly, but  GLFont::getFont(fontSize, scale) returns a Drawer proxy object that has computed the best font able to render fontSize*scale, with fontSize and scale that can be arbitrary.

So now fonts[] is only there to provide discrete Levels of Detail for the rendering and are actually independent of the scale chosen. The more fonts there are, the cleaner the rendering may be, by minimizing rescaling artifacts.
Also, because there are plenty of fonts now, they go to their own 'fonts' subdir.

The numerous commits only reflect the course of developemment, and only the squash of them make sense :)


vsonnier:
@cjcliffe  Medium (1.5x) is not bad, but Large (2x) is....well you'll see :) 
Maybe still useful in case of video projecting with a poorly defined device, with  an audience far from the screen.

cjcliffe:
@vsonnier this looks great! I'm going to make sure the new font path works with all the bundles on for each platform and then I'll merge this in.

vsonnier:
@cjcliffe  I pushed an a fix for Spectrum where frequencies steps were not properly recomputed w.r.t real font size, and also a fix to lower mhz labels w.r.t font size, so that the spectrum ticks remains visible.

vsonnier:
I've never noticed before now that the ticks marking integer values were thicker ! That was attention to detail :)

cjcliffe:
@vsonnier lol, yeah that's pretty early code too -- I'm a fan of subtle visual hints :)

cjcliffe:
@vsonnier the one I'm actually surprised nobody has caught to date is the fact that the Tuning Bar uses the Reddit up-vote / down-vote colors for tuning digits ;)

cjcliffe:
Looks good on OSX and Linux; merged to master with a few tweaks

PR #380, opened 2016-06-24
Fix for large dB values displayed with truncation on the left if > Normal zoom level, and misc.

vsonnier:
@cjcliffe  This font-related PR is made of  2 things:
- A fix for especially long dB numbers (ex: -110.6dB) being clamped on the left side if Text size > Normal.
- A "rollback" of GLPanel dynamic font size choice, and restore the previous discrete choices so that the underneath fonts are at best used in their native resolution without additional scaling. 
I did this because although the scaling works, it looks consistently ugly :(  
Maybe some (bilinear/ trilinear ) texture filtering is needed while upscaling ?

As a side note, I tried the opposite before, that is downscaling a big font and it is also ugly in just another way.

### DIFF
--- a/src/AppFrame.cpp
+++ b/src/AppFrame.cpp
@@ -103,8 +103,8 @@ AppFrame::AppFrame() :
     demodModeSelectorAdv->addChoice("QAM");
     demodModeSelectorAdv->addChoice("QPSK");
     demodModeSelectorAdv->setHelpTip("Choose advanced modulation types.");
-    demodModeSelectorAdv->SetMinSize(wxSize(44,-1));
-    demodModeSelectorAdv->SetMaxSize(wxSize(44,-1));
+    demodModeSelectorAdv->SetMinSize(wxSize(50,-1));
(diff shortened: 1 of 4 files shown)


vsonnier:
Now we need testers for Retina / 4K screens !

cjcliffe:
Looks good; didn't notice the dB issue; my NF must not have dropped that low when testing.

I think creating an a mip-mapping routine to generate the ideal mip levels on load would likely improve it for the current renderer.

But once it's all moved into shader code we'll be able to do whatever bordering / shadow / filtering tweak works best so I wouldn't put too much effort to make it work properly with the current OpenGL "immediate mode" 1.x stuff.

vsonnier:
@cjcliffe  Thanks for your answer. Of course for the texturing it is not worth your time if it is planned to rewrite the rendering ! 
As for the dB clamping, it is indeed quite easy to show up, but rather on the Audio Spectrum. As for the normal spectrum, if I zoom out at max (right-click + drag down on Waterfall part) with my RTL-SDR the min value displayed is around -105dB but it is not the kind of zoom that is actually realistic to use.

Issue #381, opened 2016-06-25
vera_sans_mono42_0.png in install rule but missing in fonts/


guruofquality:
Fails on master with make install:

```
SET (CUBICSDR_RESOURCES 
...
${PROJECT_SOURCE_DIR}/font/vera_sans_mono42_0.png
```

cjcliffe:
Fixed by @vsonnier in pull #382, cheers!

PR #382, opened 2016-06-25
Remove font 42 declaration in CMakeLists, indeed not present

vsonnier:
@cjcliffe  @guruofquality This is a fix for #381, there is really no 42pix font in the end. 
It was a leftover of the font resize dev.


cjcliffe:
aye, thanks!

PR #384, opened 2016-06-25
FIX not-initialized audio samples in mono causing out-of-bounds FFT causing non-displayed audio spectrum.

vsonnier:
@cjcliffe Most of the time when I loaded a NFM session demod, the audio spectrum didn't showed up at all. Changing for another dmod didn't changed anything, I had to stop the application and retry.

I tracked it down in ScopeVisualProcessor to a not-initialized memory for fftInData[] for mono-audio channel cases.

I also added some pedantic changes as usual:
-  Use double at some places to limit errors,
-  I C++-ified the liquid buffers into vectors, later passing them as _vector::data()_ to the liquid functions:
```cpp
std::vector<liquid_float_complex> fftInData;
std::vector<liquid_float_complex> fftOutput;
```
They should be prettier in debuggers too.

### DIFF
--- a/src/process/ScopeVisualProcessor.cpp
+++ b/src/process/ScopeVisualProcessor.cpp
@@ -10,12 +10,8 @@ ScopeVisualProcessor::ScopeVisualProcessor(): outputBuffers("ScopeVisualProcesso
 	fft_floor_ma = fft_floor_maa = 0;
     maxScopeSamples = 1024;
 #if USE_FFTW3
-    fftInData = nullptr;
-    fftwOutput = nullptr;
     fftw_plan = nullptr;
 #else
-    fftInData = nullptr;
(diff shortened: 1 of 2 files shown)


src/process/ScopeVisualProcessor.cpp
```diff
@@ -191,6 +169,7 @@ void ScopeVisualProcessor::process() {
                         fftInData[i].imag = 0;
                     } else {
                         fftInData[i].real = 0;
+                        fftInData[i].imag = 0;
```
vsonnier: The actual fix...

cjcliffe: ah, good catch thanks.

cjcliffe:
Looks good; thanks!

Issue #383, opened 2016-06-25
Sluggish performance and looking for fonts in wrong location


avsa242:
Starting with 46bf72aa0f7f94a4ad8a1f3a97808a61b0250a53, it seems CubicSDR looks for fonts under /usr/local/bin/fonts, rather than /usr/local/share/cubicsdr/fonts. Going by the messages in the terminal, the path it looks for seems to continuously grow recursively as well:
`
Font file /usr/local/bin/fonts/vera_sans_mono12.fnt does not exist?
Font file /usr/local/bin/usr/local/bin/fonts/vera_sans_mono12.fnt does not exist?
Font file /usr/local/bin/usr/local/bin/usr/local/bin/fonts/vera_sans_mono12.fnt does not exist?
`
either that or it's the same path every time, but it just happens to look recursive because of how it is output on the terminal. This causes no fonts to be rendered and the performance to be poor.

vsonnier:
@avsa242 Indeed I broke the font load by resource folder. As a temporary fix, copy the 'fonts' dir alongside the CubicSDR executable, it should load fine. 

Hopefully fixed by #385 for good.

cjcliffe:
@avsa242 latest commit to master will now install fonts to /usr/local/share/cubicsdr/fonts/ where it's expected so that the font loader doesn't try to look in /usr/local/bin/fonts/.

PR #385, opened 2016-06-25
Rewrite GLFont loading routine with correct paths computation for resource dir loading

vsonnier:
@cjcliffe @avsa242 I actually broke the loading by resource dir since I moved the fonts to 'fonts' subdir. The trouble came from that I only tested a "Windows-like" install where the fonts dir is found in the same dir as the executable. 

I so re-wrote GLFont::loadFontOnce() hopfully OK now.

### DIFF
--- a/src/util/GLFont.cpp
+++ b/src/util/GLFont.cpp
@@ -139,12 +139,12 @@ int GLFontChar::getIndex() {
     return index;
 }
 
-GLFont::GLFont(GLFontSize size, std::wstring fontFileName):
+GLFont::GLFont(GLFontSize size, std::wstring defFileName):
         lineHeight(0), base(0), imageWidth(0), imageHeight(0), loaded(false), texId(0), gcCounter(0) {
 
     fontSizeClass = size;
   
-    fontFileSource = fontFileName;
+    fontDefFileSource = defFileName;
 }
 
 GLFont::~GLFont() {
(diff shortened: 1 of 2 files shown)


cjcliffe:
I figured it might be but was going to fix it on the next bundle attempt; thanks!

cjcliffe:
Probably fixed but leaving on the list until I've tested Linux and OSX properly.

avsa242:
This seems to fix the performance problem as it isn't constantly looking for fonts when the radio is active, but it still looks for them in /usr/local/bin/fonts so won't render fonts unless manually copied there as @vsonnier suggested as a temp workaround in #383.

vsonnier:
@avsa242 Do you mean it is still falling in the curent master https://github.com/cjcliffe/CubicSDR/commit/c57478dc2ac34787d9f75427efa698a127092e09 ?

The font loading is supposed to work like this:
-  First search for fonts in resource dir: (RES_FOLDER="../share/cubicsdr/") where RES_FOLDER is a preprocessor definition given at compile time.
It means the font files (.fnt, .png) are expected here "../share/cubicsdr/fonts".
-  If a font file is not found, it is is then searched relative to the executable path, i.e it is supposed to be in "[executable path]/fonts/".

I tested both cases on Windows, and it looked OK on my side.

avsa242:
@vsonnier Yes, I pulled c57478d this morning. It still finds the fonts if I symlink /usr/local/share/cubicsdr/ to /usr/local/bin/fonts. FWIW, the installer also isn't creating a fonts subdir under /usr/local/share/cubicsdr - it puts the fonts under cubicsdr/. I've tried nuking my build dir and recreating it - same result.

vsonnier:
@avsa242 Mmm...okay, if I understand you correctly, you observe a normal situation because the fonts are no longer expected at '/usr/local/share/cubicsdr/[*.fnt|*.png]' but there: '/usr/local/share/cubicsdr/**fonts**[*.fnt|*.png]'

As for the installer, it looks like it is maintained exclusively by @cjcliffe because it doesn't appear anywhere under the GitHub repo. Indeed, the latest 'v0.20-rc2' installers are not uptodate w.r.t the fonts new location.

Can you confirm if it actually works for you for fonts located in '/usr/local/share/cubicsdr/**fonts**[*.fnt|*.png]' ?

imesg:
Works for me.

imesg:
Not only that but text size at 1.5 is... great. Just noticed. Now bookmarks.

vsonnier:
@imesg  Thanks for the confirmation. And thanks for the rest also :) 
The variable text size was primiraly intended for people with Retina / 4K class of screens (I would very much see a screenshot taken on those !) but I suprised myself using the 1.5 ratio more and more often, which is indeed very confortable on the classic 1920x1080 resolution.

imesg:
One is 1.5 and the other 2.

screenshot from 2016-06-26 11-39-36
screenshot from 2016-06-26 11-40-07

vsonnier:
@imesg Thanks ! What is your screen resolution ? Is it on OSX ?
That looks really really good !
I suppose it is > 1920x1080, because on mine x2 scale is just too big !

imesg:
It is a dell XPS 13 9350. Screen: 3200x1800 pixels. I run gentoo linux. Came with windows 10. I have a newer one that came with Ubuntu on it. I like the screen, however, it seems scaling is an issue for linux for anything not gnome (gnome is not that good either). It's coming though. You did it quickly. Let us hope others will also.

I certainly appreciate your efforts.

imesg:
Just noticed this also. The text in the manual gain section is still small maybe the same size as before. Compare both of my screenshots. The text size is the same.

vsonnier:
@imesg This is "normal" it has been left untouched because it was not simple to rise its size without affecting other graphical elements automatically... 
Maybe ask @cjcliffe for a change, he knows best about this of course.

vsonnier:
For a similar reason some other elements were just made a little bigger (demod types selection, V,S,P,M buttons, frequency tuning selectors,...) but are purposefully not affected by the Text size setting. It is a limitation of the actual implementation.

cjcliffe:
@vsonnier I can investigate tweaking the gain font; it's really just a matter of increasing the panel size that contains the text and re-balancing the sliders.. but that code was copypasta'd pretty quickly so I'll refactor it out so there isn't 30 places in there to change it ;)

cjcliffe:
Added issue #389 for manual gain font

PR #390, opened 2016-06-28
Cleaner thread termination

vsonnier:
@cjcliffe  Hello, what about another PR ?  :)
I noticed that sometimes the application hangs at exit, or at loading some session file.
This occurs not that often, but still I took another look at thread termination criteria.

I then remarked that the test *IOThread::isTerminated()* was made on the boolean IOThread::terminated, which was typically set by user to _request_ a termination, by *IOThread::terminate()*.

So it is possible that  the typical __run()__ method is not really terminated while *IOThread::isTerminated()* is returning true.

So I split the IOThread::terminated in two:
-  IOThread::stopping replacing the former IOThread::terminated, to be set the user to request termination,
-   IOThread::terminated (private) is now reflecting the thread termination, positionned automatically after __run()__ returns.

I also changed isTerminated(int waitMs = 0) to emulate a timed join() if > 0. 
In particular, I used that at application exit.

Finally, I added 'virtual' on run()/terminate() subclass redefinitions as a hint for the readers, even if it is technically not nedeed of course.

### DIFF
--- a/src/CubicSDR.cpp
+++ b/src/CubicSDR.cpp
@@ -310,6 +310,11 @@ int CubicSDR::OnExit() {
     demodVisualThread->terminate();
     t_DemodVisual->join();
 
+    //Poor man join
+    sdrPostThread->isTerminated(1000);
+    spectrumVisualThread->isTerminated(1000);
+    demodVisualThread->isTerminated(1000);
+
     delete sdrThread;
 
     delete sdrPostThread;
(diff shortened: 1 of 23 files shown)


cjcliffe:
Looks great thanks! Will give it a go and merge it in shortly 👍

cjcliffe:
All merged; just had to do the same to RigThread as well -- the Win32 and 64 libs and includes for hamlib  are available in the external/hamlib/ folder if you'd like to add it

PR #394, opened 2016-06-30
Assure center of main Waterfall+Spectrum on active demod on session load

vsonnier:
@cjcliffe Hello again!  This one is small, in order to fix something that irks me: 
When loading successive session files, the scope+waterfall demod correctly align itself on the `<active>` demod, but more often than not the Main Waterfall+Spectrum do not update w.r.t new frequency.
If I click on the Scope demod, then the Main Waterfall+Spectrum moves to present the active demod in range, often at some edge.

So this modification consist in setting explicitly the center frequency of Main Spectrum+Waterfall to be the `<active>` demod one at session loading. 

The effect is that now Main Spectrum+Waterfall is always centered on `<active>` demod when loading a session file, consistent with the demod one.

### DIFF
--- a/src/AppFrame.cpp
+++ b/src/AppFrame.cpp
@@ -1584,6 +1584,10 @@ void AppFrame::OnUnSplit(wxSplitterEvent& event)
 
 
 void AppFrame::saveSession(std::string fileName) {
+
+    //disable event processing the time of saving
+    wxEventBlocker disableEvents(this);
+
     DataTree s("cubicsdr_session");
     DataNode *header = s.rootNode()->newChild("header");
     *header->newChild("version") = std::string(CUBICSDR_VERSION);
@@ -1633,6 +1637,10 @@ void AppFrame::saveSession(std::string fileName) {
 }
 
 bool AppFrame::loadSession(std::string fileName) {
+
+    //disable event processing the time of loading
+    wxEventBlocker disableEvents(this);
+
     DataTree l;
     if (!l.LoadFromFileXML(fileName)) {
         return false;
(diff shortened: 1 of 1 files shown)


cjcliffe:
Seems to make sense; but there is a center frequency saved to session as well and it might mess with band-plan setup files where the modems are all delta-locked.  If the active modem is changing the center frequency on load then that might actually be an oversight.

I'd prefer to use the saved value and resolve any issues we're having loading it and save the current waterfall frequency/zoom state as well.

vsonnier:
@cjcliffe As far as I can see, the demods do not change the center **tunning** frequency at load, only their  displaying `InteractiveCanvas::centerFreq`. 
Indeed, unless I've got it wrong, the only times where the `InteractiveCanvas::centerFreq` sets back the **tunning** frequency is by calling the methods:
`SpectrumCanvas::moveCenterFrequency()`
 `WaterfallCanvas::updateCenterFrequency()`
which are not called at session loading, and 
by some UI events. So I think it is OK.. here is another commit FOCUS2 which explicitly disable UI events at session load and save.

vsonnier:
Adopt or discard at will, the choice is yours :) Following the flow of events acting on the settings is not very simple for sure...

cjcliffe:
I'll take a closer look at this one myself; but I still think saving the waterfall/spectrum zoom, view and frequency states is the better solution here than assuming the modem should be the focal point.

cjcliffe:
@vsonnier Check out #396 -- it's what I was thinking for the solution to this; save the view state so that it returns aligned like it was instead of loading things into the previous view state.  

I've also delayed the "loaded active modem" activation to after the center frequency and view configuration so that it shouldn't race to yank the modem into view edge on activation.

You can reproduce the original edge modem behaviour at any time by clicking the demodulator in the upper left waterfall while it's out of view below -- it's original purpose was to push it into view when you've dragged/zoomed haphazardly and want to find it :-)

vsonnier:
Closing it, @cjcliffe has made o better job with #396.

PR #395, opened 2016-07-03
thread clean 2: Assure correct termination sequence,removed NotifyQueue, and misc

vsonnier:
@cjcliffe  The quest for the perfect threading continues, following #390:
- Assure the correct termination sequence: terminate() ==> isTerminated(+ delay) ==> std::thrad:join() ==> Then only delete std::thread, IOThread objects.
- Properly nullify deleted pointers when it matters
- The "NotifyQueue" between AudioThread and DemodulatorInstance participants is actually no longer needed: the only thing it does is sending a notification on the end of IOThread::run() methods.
Later on, DemodulatorInstance::isTerminated() is called on event to examine the NotifyQueue and cleanup.

So, since now IOThread::isTerminated() is actually returning the real IOThread::run() termination and DemodulatorInstance::isTerminated() is already checking on it, the NotifyQueue mechanism double the test, and can be removed entirely.

In addition, removing NotifyQueue elmiminates another source of potential inter-thread contention.

### DIFF
--- a/src/demod/DemodulatorMgr.cpp
+++ b/src/demod/DemodulatorMgr.cpp
@@ -148,14 +148,18 @@ void DemodulatorMgr::deleteThread(DemodulatorInstance *demod) {
     if (i != demods.end()) {
         demods.erase(i);
     }
+
+    //Ask for termination
     demod->terminate();
 
+    //Do not cleanup immediatly
     demods_deleted.push_back(demod);
 }
 
-std::vector<DemodulatorInstance *> *DemodulatorMgr::getDemodulatorsAt(long long freq, int bandwidth) {
+std::vector<DemodulatorInstance *> DemodulatorMgr::getDemodulatorsAt(long long freq, int bandwidth) {
     std::lock_guard < std::recursive_mutex > lock(demods_busy);
-    std::vector<DemodulatorInstance *> *foundDemods = new std::vector<DemodulatorInstance *>();
+    
+    std::vector<DemodulatorInstance *> foundDemods;
 
     for (int i = 0, iMax = demods.size(); i < iMax; i++) {
         DemodulatorInstance *testDemod = demods[i];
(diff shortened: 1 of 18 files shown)


src/demod/DemodulatorMgr.cpp
```diff
@@ -156,9 +156,10 @@ void DemodulatorMgr::deleteThread(DemodulatorInstance *demod) {
     demods_deleted.push_back(demod);
 }
 
-std::vector<DemodulatorInstance *> *DemodulatorMgr::getDemodulatorsAt(long long freq, int bandwidth) {
```
vsonnier: No need to new() here, allocate like the others, so no need to explicitly delete later...

src/util/GLFont.cpp
```diff
@@ -400,7 +400,6 @@ void GLFont::loadFontOnce() {
         }
 
         //2) then load from memory
-        lodepng::State state;
```
vsonnier: Leftover of font size dev, removed.

vsonnier:
I think I finally pined down the hanging application problem, in the FIX commit. Turns out the SDRThread notified itself to die while dying...

cjcliffe:
This looks great; just at the cottage right now rocking my homemade Wifi Cantenna and leeching off some internet over on the shore but I'll be back home tonight for a bit and test / merge this in.

cjcliffe:
Seems significantly better so I don't think I'll hold it back; just a note if I do a quit-stress test of:

- Open with SDRPlay at about 5Mhz
- Drop about 6-8 default FM modems
- Quit after a second or two

Occasionally I get this on exit:

https://gist.github.com/cjcliffe/f4e968186a3d7043641b3e136a25b045

And have to stop it; vs. master branch where it sometimes just crashes in the audio thread and I have to stop it :-)

vsonnier:
@cjcliffe Thats how the "hang application" manifest itself on my side, with some refcount going to infinity. I found that the `SDRThread::stopping` was == false, and that the thread was going on forever. The last F  IX also reduced the problem dramatically, but still happens. The bright side is that in this case we can still attach to process and debug it... see you next time for a 'thread_clean_3' then :)

cjcliffe:
@vsonnier thanks very much for all your contributions so far; looking forward to your next commits -- have a great holiday!

PR #396, opened 2016-07-04
Save session view-state and delay loaded active modem to after frequency/view load

cjcliffe:
- Save active view state with session
- Restore view state or reset on load if none


vsonnier:
@cjcliffe This works well for me, thank you so much ! I'm closing #394 then.

PR #398, opened 2016-07-05
thread_clean_3: Multithreading AudioThread hardening, use of try_pop() and more. 

vsonnier:
@cjcliffe This is the last PR before my holidays, I promise :)
This is made of:
-  AudioThread concurrent access hardening, in particular in order to protect the static audio callback.
As a result, the audio callback is also simpler and cleaner.
-  Replace the idiom `if/while (! ThreadQueue::empty() { ThreadQueue::pop()}` which is blocking and subject to races by simply `ThreadQueue::retry_pop()` because each time it is applied here is for consuming all elements, and not to have an actual blocking waiting pop().
-  Misc : SDRThread::readStream() sometimes returns <= 0 bytes, so bail out, change some thread termination order in CubicSDR method class, (SDRThread the source of all input, is shut down first) setDevice()/stopDevice() wait or not for termination depending on we are starting for the first time or not...
-  Some comments on how things are working for (my) understanding. Yes, I do love comments. 
-  Put some TODO on things I don't understand.

Tell me if it is better or worse than before! (in terms of crashes at least)

### DIFF
--- a/src/sdr/SDRPostThread.cpp
+++ b/src/sdr/SDRPostThread.cpp
@@ -86,7 +86,7 @@ void SDRPostThread::updateActiveDemodulators() {
     nRunDemods = 0;
     
     long long centerFreq = wxGetApp().getFrequency();
-  
+
     for (demod_i = demodulators.begin(); demod_i != demodulators.end(); demod_i++) {
         DemodulatorInstance *demod = *demod_i;
         DemodulatorThreadInputQueue *demodQueue = demod->getIQInputDataPipe();
@@ -108,7 +108,9 @@ void SDRPostThread::updateActiveDemodulators() {
                 DemodulatorThreadIQData *dummyDataOut = new DemodulatorThreadIQData;
                 dummyDataOut->frequency = frequency;
                 dummyDataOut->sampleRate = sampleRate;
-                demodQueue->push(dummyDataOut);
+                if (!demodQueue->push(dummyDataOut)) {
+                    delete dummyDataOut;
+                }
             }
             
             // follow if follow mode
@@ -119,6 +121,7 @@ void SDRPostThread::updateActiveDemodulators() {
         } else if (!demod->isActive()) { // in range, activate if not activated
             demod->setActive(true);
             if (wxGetApp().getDemodMgr().getLastActiveDemodulator() == NULL) {
+
                 wxGetApp().getDemodMgr().setActiveDemodulator(demod);
             }
         }
@@ -181,8 +184,6 @@ void SDRPostThread::run() {
     iqDataOutQueue = static_cast<DemodulatorThreadInputQueue*>(getOutputQueue("IQDataOutput"));
     iqVisualQueue = static_cast<DemodulatorThreadInputQueue*>(getOutputQueue("IQVisualDataOutput"));
     iqActiveDemodVisualQueue = static_cast<DemodulatorThreadInputQueue*>(getOutputQueue("IQActiveDemodVisualDataOutput"));
-
-    iqDataInQueue->set_max_num_items(0);
     
     while (!stopping) {
         SDRThreadIQData *data_in;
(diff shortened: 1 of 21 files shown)


vsonnier:
@cjcliffe Right, I've found another source of problems: on the push() side this time, plus a real fat deadlock beetween DemodulatorMgr::terminateAll() and runing SDRPostThread, and so I'll complete this pull request until the end of the week...THEN I'll be on holidays for real :)

src/sdr/SDRPostThread.cpp
```diff
@@ -215,11 +215,14 @@ void SDRPostThread::run() {
         if (doUpdate) {
             updateActiveDemodulators();
         }
-    }
+    } //end while
     
-    if (iqVisualQueue && !iqVisualQueue->empty()) {
-        DemodulatorThreadIQData *visualDataDummy;
-        iqVisualQueue->pop(visualDataDummy);
+    //TODO: Why only 1 element was removed before ?
```
cjcliffe: visual queue has a limit of 1 element; should never be more than that in there I believe.

cjcliffe:
Looking good; I'm liking the simplification of the audio mixer.  Will give it a go here soon.

vsonnier:
Well, As I said the "hang application" problem is due to the deadloack I found, and `ThreadQueue::push()` have also a fundamental problem causing memory leaks because of its non-blocking nature and non-notification of threads waiting for pop() if > max_elements... so don't merge this pull-request yet, 'thread_clean_3.5' is coming soon.

vsonnier:
@cjcliffe TH_CLEAN_3.5 is here. First, I didn't succeed to find the source of hanging application because I lost myself between the DemodulatorInstance relationships with the other threads, in particular their correct termination order when deleting demodulators. :( 

So this final commit is just here for:
-  More AudioThread polishing,
-  More try_pop() at some missing places,
-  `ThreadQueue::push()` is non-blocking w.r.t max configured size, so check the return code and decRefCount() properly in case of failure. At least now there is no memory leaks.
-  Unsuccessfull `ThreadQueue::push()` didn't notify at all, so launch a notify to spam the waiting threads even if the queue is saturated,
-  Use a `notify_all()` in ThreadQueue to spam all waiting threads, better safe than sorry.

And that is all for now.

src/sdr/SDRPostThread.cpp
```diff
@@ -181,8 +184,6 @@ void SDRPostThread::run() {
     iqDataOutQueue = static_cast<DemodulatorThreadInputQueue*>(getOutputQueue("IQDataOutput"));
     iqVisualQueue = static_cast<DemodulatorThreadInputQueue*>(getOutputQueue("IQVisualDataOutput"));
     iqActiveDemodVisualQueue = static_cast<DemodulatorThreadInputQueue*>(getOutputQueue("IQActiveDemodVisualDataOutput"));
-
-    iqDataInQueue->set_max_num_items(0);
```
vsonnier: This one nullified the max_num_item of src/sdr/SoapySDRThread.cpp, which was set to 100, so now the check for full() now works.

src/sdr/SoapySDRThread.cpp
```diff
@@ -204,7 +204,7 @@ void SDRThread::readStream(SDRThreadIQDataQueue* iqDataOutQueue) {
         }
     }
     
-    if (n_read > 0 && !stopping) {
+    if (n_read > 0 && !stopping && !iqDataOutQueue->full()) {
```
vsonnier: So, purely discard the incoming readStream()-ed data if the output queue is actually full (max is set to 100)

cjcliffe:
Looking good! Going to give it a go here on OSX

cjcliffe:
Working well here for OSX and Linux; going to merge it in -- thanks again for your dedicated efforts!

PR #416, opened 2016-07-28
Crash with non-empty ModemArgInfoList (in FM Stereo)

vsonnier:
@cjcliffe Hello, back from holidays ! 
Since FM deemphasis has been added, we have actually non-empty ModemArgInfoList showing up in regular build (i.e. excluding Digital Lab modems). When loading alternatively session files with FMS and NFM modems, a crash occurr after 2-3 session loads max.
VisualStudio reported some horrible 'corrupted memory' somewhere in 
``````cpp
void AppFrame::updateModemProperties(ModemArgInfoList args) {
    newModemArgs = args; ==> something wrong goes there, in assignement/deletion/ of std::vector<ModemArgInfo>
....
`````` 
So since the `std::vector<ModemArgInfo>` do not contain any fancy structures, the problem couldn't happen from wrong assignment code. On the other hand, AppFrame::updateModemProperties() is called by any DemodulatorWorkerThread, so it may be an unprotected concurrent access problem.

Luckily AppFrame::newModemArgs looks actually unsed because settings are read per-Modem in DemodulatorInstance::getModemArgs() if I'm right, so I simply removed AppFrame::newModemArgs entirely and the crashes actually disappeared.

### DIFF
--- a/src/AppFrame.cpp
+++ b/src/AppFrame.cpp
@@ -1496,8 +1496,12 @@ void AppFrame::OnIdle(wxIdleEvent& event) {
     demod = wxGetApp().getDemodMgr().getLastActiveDemodulator();
     
     if (modemPropertiesUpdated.load() && demod && demod->isModemInitialized()) {
-        modemProps->initProperties(demod->getModemArgs());
+        
+        //reset notification flag
         modemPropertiesUpdated.store(false);
+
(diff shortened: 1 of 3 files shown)


cjcliffe:
@vsonnier good catch thanks; looks like that bit of dead code was from some incomplete changes from awhile ago.

PR #422, opened 2016-08-12
Squelch Improvements

cjcliffe:
Improvements:
- Basic auto-normalization of the signal range with moving floor/ceiling targets; allows more detailed control of squelch trigger instead of most weak signals just being floored with a wiggle.

### DIFF
--- a/src/demod/DemodulatorThread.cpp
+++ b/src/demod/DemodulatorThread.cpp
@@ -15,7 +15,7 @@
 
 DemodulatorThread::DemodulatorThread(DemodulatorInstance *parent) 
     : IOThread(), outputBuffers("DemodulatorThreadBuffers"), squelchLevel(-100), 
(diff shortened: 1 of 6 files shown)


src/demod/DemodulatorThread.cpp
```diff
@@ -114,6 +114,33 @@ void DemodulatorThread::run() {
...
+        
+        if (sl+1.0f > sc) {
+            sc = sl+1.0f;
+        }
+
+        if ((sf+2.0f) > sc) {
+            sc = sf+2.0f;
+        }
+        
+        sc -= (sc - (currentSignalLevel + 2.0f)) * sampleTime * 0.15f;
```
cjcliffe: - [ ] TODO: define these as the balance parameters

cjcliffe:
Working better so far but squelch jump is still sketchy even with good squelch settings; need to add some basic logic for it.

Issue #423, opened 2016-08-13
New squelch algorithm not working maybe because IQ power vs. Audio power comparisons


vsonnier:
@cjcliffe Hi ! I've tested the new squelch management, but it doesn't work well for me.
A video to demonstrate the problem:
https://www.youtube.com/watch?v=ZqkiRoE0NNw

This is NFM listening, quite powerful against the noise floor (>-15dB with a floor at -85dB).
The moment the transmission stops, a very loud "audio noise" > "normal voice" actually prevents the squelch to work.

I think the trouble lies in the fact that we are comparing apples and oranges when we compare audio dBs and IQ dBs to select the squelch level in DemodulatorThread.cpp (161):
````cpp
currentSignalLevel = iqSignalLevel>audioSignalLevel?iqSignalLevel:audioSignalLevel;
````    
I can [see](https://github.com/cjcliffe/CubicSDR/commit/6ea2b5b8e81d824ee42a288f5d60a8bd147c000e) that you introduce the change on purpose for likely a good reason, but I don't think those quantities can be compared. In this use case above, an "audio noise" has much more apparent power than the corresponding IQ power, so nullifying the squelch..

So I tried to only use the IQData like before the squelch changes to see what it does:
``````cpp
//TEST:
currentSignalLevel = iqSignalLevel;
``````
You see the result in the following video:
https://www.youtube.com/watch?v=q-yb0D2modw

The squelch is working perfectly there, and much better than before your squelch improvement branch!  This also works great with:
-  POCSAG the old audio-encoded Pager protocol,  
-  DMR digital voice, piping audio to DMR+ decoder, using a squelch level of zero.

All in all, I don't see how IQ power and Audio power can really be compared to each other, their nature is different...

cjcliffe:
It's been somewhat balanced here but it was a struggle to get it to work well with POCSAG / DMR here too; I have a feeling if I change my gains it won't work as well.

I'll change it for now so that FM based uses the I/Q signal and AM based uses the audio signal -- better AM, LSB and USB responses were the primary reason for changing.

cjcliffe:
@vsonnier should be patched up now.

vsonnier:
@cjcliffe It works rater well for AM, tested 131-138Mhz, 8khz large plane communications: Signal a drowned into noise, and do not occur very often. Still, the squelch managed to mute most of the noise without masking the real transmissions... it is certainly better than before. 
I've added some small simplifications to it with #425.

[...]

Issue #411, opened 2016-07-26
'No SoapySDR module found' error with custom built Windows builds 


vsonnier:
@cjcliffe Hello ! I'm trying to compile the master for Win64, reusing all the other v0.20-rc4-win64 release binaries, which is what I do to use/test the master usually. What I'm getting now is this:
![no_modulesfound](https://cloud.githubusercontent.com/assets/5152742/17139984/ff9861ca-5346-11e6-8644-634c04c9b5ee.jpg)
poping up like mad.
 
The CubicSDR exe from https://github.com/cjcliffe/CubicSDR/releases/tag/0.2.0-beta-rc5 actually runs fine, though. With some build specific settings maybe ?

A workaround is to provide the module path on command line explicitly like `...\CubicSDR.exe -m ./modules/`

I actually started to observe this when attempting to compile the SoapySDR v0.5+ some time ago, but since it is built-in now I can no longer dodge the problem.

cjcliffe:
Hmm, for local testing in Windows I use the active SoapySDR paths -- Just make sure the soapy module install prefixes are set to C:\Program Files\SoapySDR\ and the SoapySDR bin folder is in your path as well as the SoapySDR environment variable should be C:\Program Files\SoapySDR\ (this stuff changes if you installed pothos)  

From there you copy any relevant bins for the modules into the SoapySDR bins folder.

guruofquality:
> I actually started to observe this when attempting to compile the SoapySDR v0.5+ some time ago, but since it is built-in now I can no longer dodge the problem. 

Perhaps this explains it? So, I havent tagged 0.5.1 yet, but there was a change to the modules directory to address a multi-version install on debian packaging issue... more details: https://github.com/pothosware/SoapySDR/pull/85

I wasn’t sure how this would manifest because I thought cubic used its own loader, but I can add back the original search path. Its just looking in INSTALL_ROOT/lib/SoapySDR/modules.0.5-3 But any recompiled modules should install to that location.

vsonnier:
[...]

PR #425, opened 2016-08-13
Small cleanup squelsh and misc

vsonnier:
@cjcliffe it is a small simplification of the current squelch, including #423 changes:
__SMALL_SQUELCH_CLEANUP:__
-  in DemodulatorThread::abMagnitude(), remove the alpha-beta filtering by the real norm computation using sqrt() like everywhere else in code. On desktop CPU sqrt() is a free intrinsic for ages, no need to DSP approximations,
-  In squelch code, use double instead of floats to keep some precision when accumulating or computing magnitudes,

__WINDOWS_NO_SECONDARY_CONSOLE:__
This in unrelated to squelch, but I noticed that not-allocating an explicit Console on Windows made the application less hang prone when loading / reloading demodulators (through sessions, typically)
 
In addition, following my investigation about #411, I noticed that std::cout traces actually shows up when running the executable in a MINSYS2 environment like Cygwin or Git Bash for Windows, so there is no need to allocate an had-hoc Console to have debug traces.
This commit is just about disabling the Console allocation code entirely by an `#if 0` block.

### DIFF
--- a/src/CubicSDR.cpp
+++ b/src/CubicSDR.cpp
@@ -165,8 +165,11 @@ bool CubicSDR::OnInit() {
         return false;
     }
 
-    //	// console output for Windows: available in DEBUG or in case of ENABLE_DIGITAL_LAB
-#if (defined(WIN32) && (defined(ENABLE_DIGITAL_LAB) || defined(_DEBUG)))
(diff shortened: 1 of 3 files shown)


cjcliffe:
all looks great, thanks!

PR #436, opened 2016-09-02
use the system RtAudio on Linux

stefantalpalaru:
Tested on Gentoo ~amd64 with rtaudio-4.1.2 . It will use whatever backends were enabled in the system-wide RtAudio library, auto-detecting the required defines, includes and libraries thanks to pkg-config.


cjcliffe:
@stefantalpalaru this looks good; will try updating the flatpak and appimagekit builds to use it and merge if all works out.

cjcliffe:
Hey @stefantalpalaru -- if you can add a boolean build flag to the CMakeLists.txt to make system RtAudio optional and revert the built-in version and flags then I'll merge this in.  It's likely to cause problems with Windows, OSX and users currently building from master as it is now.  Thanks!

cjcliffe:
@stefantalpalaru update looks good; will try it here soon and merge -- I like the idea of using a macro to group the dependencies -- I may use that to clean up some other platform-specific cmake logic.

Cheers!

PR #437, opened 2016-09-02
FIX: saved or created session files in v0.2.0 crashes on reload

vsonnier:
@cjcliffe This is the first bug report of version v0.2.0, hittting exactly Murphy's law:

Either load a session file, and resave it, or create a new session file in v0.2.0, then finally try to reload it: 
BOOM.

The problem lies in the XML tag 
`<version>0.2.0</version>` 
where '0.2.0' is interpreted as DATA_FLOAT by DataTree on XML reload, but when it is actually retreived in code is interpreted as a string ==> access violation / exception... The case didn't arise before because '0.2.0-beta-rcX' were recognized as a string proper :)

The only clean solution I found is to actually store and load `<version>` as a std::wstring just like `<user_label>` so it can really contain any kind of characters without the risk of being interpreted other than a DATA_WSTRING throughout.

I also added a small safety net so that older session files can still be loaded in v0.2.0 while their version 'encoding' has become incompatible. Re-save them afterwards and everything should be clean.

### DIFF
--- a/src/AppFrame.cpp
+++ b/src/AppFrame.cpp
@@ -1633,7 +1633,9 @@ void AppFrame::OnUnSplit(wxSplitterEvent& event)
 void AppFrame::saveSession(std::string fileName) {
     DataTree s("cubicsdr_session");
     DataNode *header = s.rootNode()->newChild("header");
-    *header->newChild("version") = std::string(CUBICSDR_VERSION);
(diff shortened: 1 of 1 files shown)


cjcliffe:
@vsonnier lol; I hadn't considered this -- good catch was about to debug it :)

Will update 0.2.0 with this patch.

Issue #483, opened 2017-01-03
Compilation errors on OS X with the new version


mehdideveloper:
Hi
After checking out the new version, I've got multiple compilation errors (It was OK the last time I checked, about 1 month ago)

Error log: http://pastebin.com/JKPNREiA

starvald:
no good for me either on ubuntu 16.04 x64.

same errors.

vsonnier:
Confirmed also on Win64. Problem should be fixed by #484.

cjcliffe:
Thanks all for getting this reported and fixed quickly! 👍

PR #484, opened 2017-01-03
FIX: wrong directive placement prevent building without HAMLIB

vsonnier:
@cjcliffe I wish you a Happy New Year ! I've started to test your Bookmarks enhancement, it looks really good:)
Still, I've encountered some problems on Win64   
- This fixes the wrong placement of some declared variables placed within #if USE_HAMLIB / #endif, preventing compilation without HAMLIB enabled. Problem was reported in #483 . 

I've got another problem: Demod Spectrum and Waterfall remains empty, I've not started to investigate this one yet (Reproducible with your Win64 021alpha build) I suspect the DEMOD_VISIBLE option is at play here... Since you can't have missed this on OSX, I would say it is a non-initialized flag beahaving differently on Windows somewhere. Image: 
2017-01-03_205850_small_cr

### DIFF
--- a/src/AppConfig.h
+++ b/src/AppConfig.h
@@ -179,9 +179,10 @@ private:
     std::atomic<float> spectrumAvgSpeed, mainSplit, visSplit, bookmarkSplit;
     std::atomic_int dbOffset;
     std::vector<SDRManualDef> manualDevices;
+    std::atomic_bool bookmarksVisible;
 #if USE_HAMLIB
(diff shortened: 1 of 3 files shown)


cjcliffe:
Good catch thanks; hadn't built without hamlib recently or I might have caught that.

Other one still stumping me as well; if the demod visible flag was off it should be omitting the entire modem spectrum and waterfall panel and not just the waterfall...

PR #490, opened 2017-01-08
FIX: missing demod displays due to wrong input/queue setups order

vsonnier:
@cjcliffe Hello ! I caught the missing demod wf/spectrum problem. It was cased by a change in initialization order between setInput/OutputQueue in CubicSDR.cpp vs. v0.20. We have now in the current version:
````cpp
t_PostSDR = new std::thread(&SDRPostThread::threadMain, sdrPostThread);
...
sdrPostThread->setOutputQueue("IQActiveDemodVisualDataOutput", pipeDemodIQVisualData);
````

This is indeed wrong, because like most of long-running threads, the loop is crafted like this:
 ````cpp
run() {
   // Retreive input/output queues pointers once... 
  while(!stopping) {
   // work with those pointers...
   }
}
````
This is OK as long as the input/output setup is completely done BEFORE the thread starts, which is not he case above.

So here are the changes :
-  CubicSDR.cpp: changed order so that threads are all started AFTER input/output plumbing,
-  IOThread: We may want to change queues dynamically, or just re-read by getInput/Output, (like in DemodulatorInstance) so the set/get must be cleanly protected against concurrent access.
-  AppFrame.cpp: completly useless ifs re-formating :)

I've checked the other usage patterns of set/get queues vs. threads, they look fine.

### DIFF
--- a/src/AppFrame.cpp
+++ b/src/AppFrame.cpp
@@ -281,7 +281,7 @@ AppFrame::AppFrame() :
 
 //    vbox->Add(demodTray, 12, wxEXPAND | wxALL, 0);
 //    vbox->AddSpacer(1);
-    bookmarkSplitter = new wxSplitterWindow( mainSplitter, wxID_BM_SPLITTER, wxDefaultPosition, wxDefaultSize, wxSP_3DSASH | wxSP_LIVE_UPDATE );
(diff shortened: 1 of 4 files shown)


cjcliffe:
@vsonnier awesome thanks; working well here.

Issue #393, opened 2016-06-30
Compile Issue


heath33:
Compile failing at:

`_/usr/local/src/CubicSDR/src/modules/modem/analog/ModemFM.h:19:5: error: ‘freqdem’ does not name a type
     freqdem demodFM;
     ^_`

I saw the other post about the different versions of liquid-dsp.  I can install liquid-dsp from the tarball, but it will not install from the github source (as per the instructions on CubicSDR.)  So I have liquid-dsp installed, but CubicSDR fails, I can't install liquid-dsp per guidance, it fails here

`In file included from src/dotprod/src/dotprod_cccf.mmx.c:49:0:
src/dotprod/src/dotprod_cccf.mmx.c: In function ‘dotprod_cccf_execute_mmx’:
/usr/lib/gcc/i686-linux-gnu/5/include/pmmintrin.h:50:1: error: inlining failed in call to always_inline ‘_mm_addsub_ps’: target specific option mismatch
 _mm_addsub_ps (__m128 __X, __m128 __Y)`

The system I am trying to install this on should be up to date with all dependencies met.
System:    Host: Gateway Kernel: 4.6.0-1-686-pae i686 (32 bit) Desktop: Cinnamon 3.0.4
           Distro: Debian GNU/Linux stretch/sid
Machine:   System: Gateway (portable) product: MT6705 v: 3408385R
           Mobo: Gateway model: N/A v: 77.10 Bios: Phoenix v: 77.12 date: 04/18/2007
Battery    BAT1: charge: 27.4 Wh 98.4% condition: 27.9/48.0 Wh (58%)
CPU:       Dual core Intel T2060 (-MCP-) cache: 1024 KB 
           clock speeds: max: 1600 MHz 1: 1333 MHz 2: 1067 MHz
Graphics:  Card: Intel Mobile 945GM/GMS 943/940GML Express Integrated Graphics Controller
           Display Server: X.Org 1.18.3 drivers: intel (unloaded: fbdev,vesa) Resolution: 1280x800@60.00hz
           GLX Renderer: Mesa DRI Intel 945GM x86/MMX/SSE2 GLX Version: 2.1 Mesa 11.2.2

cjcliffe:
@heath33 Ahoy -- You'll need a newer liquid-dsp somehow; older versions of liquid-dsp had an integrated FM modem but there's now a separate modulator and demodulator.  There's other compile errors past that you'll get with an older version as well.

[...]

Issue #338, opened 2016-04-26
Bug: phantom stations on the screen


dmitryelj:
There are 2 cases, I've found:
- top-left window - at the edge of the screen sometimes appears the signals, that are disappeared when user scroll them to the center or tap on them
- main spectrum window - when I zoom in (with mouse wheel) at some zoom level appear new stations that are really not exist (when clicking on them no sound and nothing at the top-left view).

Mac OS, receiver: SDRPlay, S/R: 2M (sometimes 5M), frequencies: 100-300KHz.

cjcliffe:
@dmitryelj CubicSDR utilizes decimation on the main waterfall which can introduce aliases -- it's not really a bug more of an aliasing / optimization artefact which I may be able to reduce in the future with some filtering.  Changing the center frequency while zoomed can sometimes knock them out of the way or just zoom in/out a bit more until they flatten out.

There's a similar known issue; if you go hunting in the empty spaces between signals (especially in the WBFM bands) the polyphase filterbank channelizer will sometimes create signal aliases you can actually hear that aren't visible on the main waterfall but will show in the top-left view.  Those signals are just wrapped from shifting the baseband of the channel the signal falls into -- I posted a detailed description about the issue here: #295

dmitryelj:
![phantom-1](https://cloud.githubusercontent.com/assets/18378056/14870836/6d72bf52-0cdd-11e6-8eeb-ae2be4569690.png)
![phantom-2](https://cloud.githubusercontent.com/assets/18378056/14870837/6d73b90c-0cdd-11e6-9c17-7288cac5f648.png)

I've attached 2 screenshots.
Thats really annoing because this signals have big levels, and it's hard to differentiate them from real, I need to play with frequency and zoom to determine if them disappear or not.

cjcliffe:
@dmitryelj I agree it's a bit annoying and I do hope to improve it so I'll mark this down as an enhancement to look at for 0.2.x -- I know there's more I can do with filtering and improving the decimation transitions at least.

vsonnier:
@cjcliffe Hello ! Following the issue #556 where the ghost/issues are poping up like the real bothering ones  for otherwise enthusiastic users, I re-read the whole set of issues and started experimenting from my understanding of them.
As a result, I propose the following user-selectable 'performance' levels for those who want to burn more CPU:

Perf. level  |  Channelizer size | internal FFT size | Remarks
----------- | ------------------- | ------------------| -----------
[...]

PR #497, opened 2017-01-14
Liquid dsp v1.30 upgrade (for Windows)

vsonnier:
@cjcliffe Hello ! This is an attempt to upgrade liquid-dsp to v1.30 recently released.
The work consisted in rebuilding the libs for Win32/Win64:
-  I modified a makefile initially generated on a Linux box.

- Then, thanks to information collected on liquid-dsp current Github issues (building on MinGW, pactching the makefile and so on) I was able to build Windows libraries using the Msys2 environnement. (gcc 6.2)

- I have deleted external/liquid-dsp/msvc/[32/64]/libliquid.exp files, they are of no use for compiling with Visual Studio.
- Win32 dll now needs both libgcc_s_dw2-1.dll and libwinpthread-1.dll, so I have added it to externals and changed CMakeLists.txt.
- I have upgraded Visual C++ 2015 runtimes to their most recent version.
- I have also removed the LIQUID_EXPERIMENTAL of Digital Lab, because now it is standard :)

Despite the numerous API changes in liquid, CubicSDR compile without modification because apparently it does not use the impacted APIs.

As far as I tested, both Win32 / Win64 seem to behave as before.
 
I had the idea of upgrading because of the amount of improvements shown from v1.20 version: For instance in re-samplers, so that #338 is maybe improved ? (I don't know)

At least v1.30 has a richer API that may be usefull for future devs.

For compilling on Linux, you'll need to patch the generated makefile as reported in #393 and linked liquid-dsp issues, by adding  ````-msse3 ```` to  ````CONFIG_CFLAGS````. 

I've not touched OSX, so it is terra incognita.


cjcliffe:
@vsonnier I believe Windows was the only platform really left behind at the moment so this is perfect -- the ones you replaced were the original liquid-dsp DLLs downloaded from the original incarnation of the liquid-dsp site during initial development -- which are no longer available online :)

I'm using the latest liquid-dsp sources from master on OSX and Linux so I assume there shouldn't be any conflicts there whatsoever.

PR #499, opened 2017-01-21
Device Sample Rate improvements: UI, rates handled as long, misc... 

vsonnier:
@cjcliffe Good day ! This introduce small changes in handling of device sample rate, mainly to help with UI freindlness. It fixes in particular:
-  Select Manual Entry, set a frequency. The radio button is now checked. Then, if you want to modify it again, you have to select another choice and get back to Manual Entry.
-  The sample_rate stored in session file is not restored at session load, another choice is made even if the saved one is perfectly compatible with the current device properties.
-  When selecting a sample rate in the SDR Device dialog and pressing Start, the AppFrame is not updated back, in the menus in particular, so that the displayed sample rate do not match the one actually applied. Take care now that AppFrame is updated in this case where it is not the origin of the changes. 
-  Following the path of  device ````setSampleRate()/getSampleRate()```` in the whole code, I saw that some were int, others long and long long mixed together. Mostly harmless unless going beyond 2**31 Hz, but well... I turned the int to long or long long.
      
In order to fix the first point, I had to modify the Sample Rate Menu a bit, because a single Radio Button couldn't handle the change (clicking on a already checked radio button triggers no event !)

So, I've added  a normal "Manual Entry ..." item dedicated to the update of  manual sample rate through the dialog, and modified the existing radio button into "Manual: [freq] Mhz" displaying the latest manual sample rate entered by the operator.
I think it works quite well you'll see.

Finally I couldn't stop myself adding some pedantic traces and coments here and there.

### DIFF
--- a/src/AppFrame.cpp
+++ b/src/AppFrame.cpp
@@ -709,9 +709,13 @@ void AppFrame::initDeviceParams(SDRDeviceInfo *devInfo) {
     deviceChanged.store(true);
 }
 
+void AppFrame::notifyDeviceChanged() {
+    deviceChanged.store(true);
+}
+
 void AppFrame::updateDeviceParams() {
     
-    if (!deviceChanged.load()) {
+    if (!deviceChanged.load() || devInfo == nullptr) {
         return;
     }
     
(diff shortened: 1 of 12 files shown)


cjcliffe:
@vsonnier excellent work as always; thanks again!  I'll keep in mind to start commenting the existing code myself too to help make sense of things -- cheers!

vsonnier:
@cjcliffe Thanks, the pleasure is mine :+1:

PR #501, opened 2017-01-27
UI fixes: Theme set/change, B shortcut, add icon to SDR Devices dialog.

vsonnier:
@cjcliffe Hello again, here is another batch of UI fixes:

-  First the real problem: text of bookmark buttons is black, on a very dark background, so almost invisible. Turns out the color theme was not applied to such buttons, which also leads to...
-  Fixing complete color theme setup: lack of (instant) repaint of Digital lab modem buttons, no (instant) repaint of modem properties dialog.  Every control is now correctly repaint immediatly with the new theme colors.
-  As cleanup, isolate the Display menu related code in its own function, because ``AppFrame::OnMenu()`` is large enough by now.

One limitiation of theme color application I won't try to overcome: ``wxButton`` colors can't be changed after construction (at least on Windows). 
So if the color theme is changed, the button colors remain the same no matter what. Fortunately since selecting another bookmark tree element destroys and re-creates the buttons, a color refresh occurs then.  

- 'B' (dB display toggle) is supposed to activate on Spectrum mouse hover, not Waterfall. Wonders for quite some time how the machinery worked, discovers what ``MouseTracker`` was  about :+1:  
So, I plugged the redispatch of events correctly to ``SpectrumCanvas`` and ``ScopeCanvas``, which were missing, then moved the code for 'B' key from ``WaterfallCanvas`` to ``SpectrumCanvas`` where it belongs.
-  Make the Audio spectrum have its own, independent 'B' handling to disable its dB display independently from the Spectrum one.
-  Give the SDR Devices dialog a proper icon as the main application has. 

That's is all for now, i wish you a nice week-end !

### DIFF
--- a/src/AppFrame.cpp
+++ b/src/AppFrame.cpp
@@ -207,7 +207,7 @@ AppFrame::AppFrame() :
 
 #if CUBICSDR_ENABLE_VIEW_SCOPE
     scopeCanvas = new ScopeCanvas(demodPanel, attribList);
-    scopeCanvas->setHelpTip("Audio Visuals, drag left/right to toggle Scope or Spectrum.");
+    scopeCanvas->setHelpTip("Audio Visuals, drag left/right to toggle Scope or Spectrum, 'B' to toggle decibels display.");
     scopeCanvas->SetMinSize(wxSize(128,-1));
     demodScopeTray->Add(scopeCanvas, 8, wxEXPAND | wxALL, 0);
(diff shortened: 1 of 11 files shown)


cjcliffe:
@vsonnier many thanks this is great -- I've added you as a collaborator so you'll be able to manage a bit more and pull the trivial fixes without having to wait for me :1st_place_medal:

vsonnier:
@cjcliffe  Thank you very much !

PR #502, opened 2017-01-29
CLEANUP: some compiler and static analysis warnings

vsonnier:
@cjcliffe This a pass over some compiler warnings: type conversions, uninitialized variables...
No problem was found, but that allows to reduce the compiler warning clutter so that real errors will show up more clearly in the future. 
For type conversions, applying explicit casts make those conversions visible in code which is often a good thing.
Hope I don't break anything, ssince I'll merge that by myself :)


vsonnier:
@cjcliffe Alea jacta est

cjcliffe:
@vsonnier nice work, thanks again 👍

PR #510, opened 2017-02-05
Waterfall fixes and improvements

vsonnier:
@cjcliffe My advantures in the inner workings of CubicSDR continues... This time, VisualProcessor and FFTDataDistributor. In FFTDataDistributor, I caught 2 bugs (see code annotations) one about loosing samples (BUGFIX1) and the other a frozen waterfall if conditions are met (BUGFIX2).

I've also added OPTION1 which consist in setting 2 times many waterfall lines, both in demod and main waterfall. This gives precise and pretty details, and do not seem to eat more CPU (tested on high end Core i7 mobile from 5 years ago with integrated graphics), and of course the main waterfall is 2 times slower  for a given line-per-sec setting.
For this one I'll let you decide if you want it or not, so I'm not going to push this PR by myself.

### DIFF
--- a/src/CubicSDRDefs.h
+++ b/src/CubicSDRDefs.h
@@ -31,7 +31,15 @@ const char filePathSeparator =
 #define BUF_SIZE (16384*6)
 
 #define DEFAULT_SAMPLE_RATE 2500000
+
+//
 #define DEFAULT_FFT_SIZE 2048
+#define DEFAULT_DMOD_FFT_SIZE (DEFAULT_FFT_SIZE / 2)
+#define DEFAULT_SCOPE_FFT_SIZE (DEFAULT_FFT_SIZE / 2)
+
+//Both must be a power of 2 to prevent terrible OpenGL performance.
+#define DEFAULT_MAIN_WATERFALL_LINES_NB 512
+#define DEFAULT_DEMOD_WATERFALL_LINES_NB 128
 
 #define DEFAULT_DEMOD_TYPE "FM"
 #define DEFAULT_DEMOD_BW 200000
@@ -43,3 +51,5 @@ const char filePathSeparator =
 #define MANUAL_SAMPLE_RATE_MIN 2000000 // 2MHz
 #define MANUAL_SAMPLE_RATE_MAX 200000000 // 200MHz (We are 2017+ after all)
 
+//Represents the amount of time to process in the FFT distributor. 
+#define FFT_DISTRIBUTOR_BUFFER_IN_SECONDS 0.250
--- a/src/CubicSDRDefs.h
+++ b/src/CubicSDRDefs.h
@@ -38,13 +38,18 @@ const char filePathSeparator =
 #define DEFAULT_SCOPE_FFT_SIZE (DEFAULT_FFT_SIZE / 2)
 
 //Both must be a power of 2 to prevent terrible OpenGL performance.
-#define DEFAULT_MAIN_WATERFALL_LINES_NB 512
-#define DEFAULT_DEMOD_WATERFALL_LINES_NB 128
+#define DEFAULT_MAIN_WATERFALL_LINES_NB 1024
+#define DEFAULT_DEMOD_WATERFALL_LINES_NB 256
 
 #define DEFAULT_DEMOD_TYPE "FM"
 #define DEFAULT_DEMOD_BW 200000
 
 #define DEFAULT_WATERFALL_LPS 30
+
+//Dmod waterfall lines per second is adjusted 
+//so that the whole demod waterfall show DEMOD_WATERFALL_DURATION_IN_SECONDS
+//seconds.
+#define DEMOD_WATERFALL_DURATION_IN_SECONDS 4.0
 
 #define CHANNELIZER_RATE_MAX 500000
 
--- a/src/process/FFTDataDistributor.cpp
+++ b/src/process/FFTDataDistributor.cpp
@@ -2,13 +2,15 @@
 // SPDX-License-Identifier: GPL-2.0+
 
 #include "FFTDataDistributor.h"
+#include <algorithm>
 
 FFTDataDistributor::FFTDataDistributor() : outputBuffers("FFTDataDistributorBuffers"), fftSize(DEFAULT_FFT_SIZE), linesPerSecond(DEFAULT_WATERFALL_LPS), lineRateAccum(0.0) {
 
 }
 
-void FFTDataDistributor::setFFTSize(unsigned int fftSize) {
-	this->fftSize = fftSize;
+void FFTDataDistributor::setFFTSize(unsigned int size) {
+	 
+    fftSize.store(size);
 }
 
 void FFTDataDistributor::setLinesPerSecond(unsigned int lines) {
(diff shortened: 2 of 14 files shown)


src/CubicSDRDefs.h
```diff
@@ -31,7 +31,15 @@ const char filePathSeparator =
 #define BUF_SIZE (16384*6)
 
 #define DEFAULT_SAMPLE_RATE 2500000
+
+//
 #define DEFAULT_FFT_SIZE 2048
+#define DEFAULT_DMOD_FFT_SIZE (DEFAULT_FFT_SIZE / 2)
```
vsonnier: define all the things !

src/process/FFTDataDistributor.cpp
```diff
@@ -29,25 +31,50 @@ void FFTDataDistributor::process() {
 		input->pop(inp);
 
 		if (inp) {
+            //Settings have changed, set new values and dump all previous samples stored in inputBuffer: 
 			if (inputBuffer.sampleRate != inp->sampleRate || inputBuffer.frequency != inp->frequency) {
-                
-                bufferMax = inp->sampleRate / 4;
+
+                //bufferMax must be at least fftSize (+ margin), else the waterfall get frozen, because no longer updated.
+                bufferMax = std::max((size_t)(inp->sampleRate * FFT_DISTRIBUTOR_BUFFER_IN_SECONDS), (size_t)(1.2 * fftSize.load()));
```
vsonnier: BUGFIX2 part1 : if bufferMax < fftSize, waterfall get frozen in the following situation : a low sampling rate and big waterfall zoom  which needs a big fftSize.

src/process/FFTDataDistributor.cpp
```diff
@@ -29,25 +31,50 @@ void FFTDataDistributor::process() {
...
                 bufferOffset = 0;
-                
+                bufferedItems = 0;
 				inputBuffer.sampleRate = inp->sampleRate;
 				inputBuffer.frequency = inp->frequency;
                 inputBuffer.data.resize(bufferMax);
 			}
+
+            //adjust (bufferMax ; inputBuffer.data) in case of FFT size change only.
+            if (bufferMax < (size_t)(1.2 * fftSize.load())) {
```
vsonnier: BUGFIX2 part2: inputBuffer.data must adapt to the chnages of fftSize only.

src/process/FFTDataDistributor.cpp
```diff
@@ -29,25 +31,50 @@ void FFTDataDistributor::process() {
...
+                inputBuffer.data.resize(bufferMax);
+            }
+
+            size_t nbSamplesToAdd = inp->data.size();
+
+            //No room left in inputBuffer.data to accept inp->data.size() more samples.
+            //so make room by sliding left of bufferOffset, which is fine because 
+            //those samples has already been processed.
             if ((bufferOffset + bufferedItems + inp->data.size()) > bufferMax) {
                 memmove(&inputBuffer.data[0], &inputBuffer.data[bufferOffset], bufferedItems*sizeof(liquid_float_complex));
```
vsonnier: BUGFIX1: Previously, after the memmove (compacting) the memcpy was wrongly in a else branch, so no copying occured and the samples were lost.

src/process/FFTDataDistributor.cpp
```diff
@@ -29,25 +31,50 @@ void FFTDataDistributor::process() {
...
                 bufferOffset = 0;
-            } else {
-                memcpy(&inputBuffer.data[bufferOffset+bufferedItems],&inp->data[0],inp->data.size()*sizeof(liquid_float_complex));
-                bufferedItems += inp->data.size();
+                //if there are too much samples, we may even overflow !
+                //as a fallback strategy, drop the last incomming new samples not fitting in inputBuffer.data.
+                if (bufferedItems + inp->data.size() > bufferMax) {
```
vsonnier: BUGFIX1 : after compacting, limit the copy to bufferMax anyway.

src/process/VisualProcessor.h
```diff
@@ -78,37 +81,51 @@ class VisualProcessor {
...
-        // process inputs to output
-        // distribute(output);
-    }
-
-    void distribute(OutputDataType *output) {
-        // distribute outputs
+    // derived class must implement a  process() interface
+    //where typically 'input' data is consummed, procerssed, and then dispatched
+    //with distribute() to all 'outputs'.
+    virtual void process() = 0;
```
vsonnier: make process() a true interface to force concrete implementations in derived classes.

src/process/FFTDataDistributor.cpp
```diff
@@ -56,12 +83,14 @@ void FFTDataDistributor::process() {
 		// number of lines in input
 		double inputLines = (double)bufferedItems / (double)fftSize;
 
-		// ratio required to achieve the desired rate
+		// ratio required to achieve the desired rate:
+        // it means we can achieive 'lineRateStep' times the target linesPerSecond.
+        // < 1 means we cannot reach it by lack of samples.
```
vsonnier: @cjcliffe My understanding of the principles:
-  if lineRateStep < 1, we are in sample famine, so consume all,
 else
-  Find a way to only distribute() at most the required line-per second rate, by eliminating some samples,

Question 1: Is it correct ? I don't understand the code below, I admit...
Question 2: Why not using a liquid resampler to adapt the number of samples ?

cjcliffe: @viraptor Ahoy, 

1. That sounds about right; it's main purpose was to grab the frame of samples that best lines up with the actual timing of the waterfall.
2. Additional re-sampling seemed to create artifacts / aliases and requires more CPU; all the waterfall needs is a clean frame that's the correct size at the right time interval to keep things looking nice and from over/under flowing; the waterfall processor only re-samples using 1/2 decimation stages as needed and in-between it's just interpolating the data for visual display, which is much cheaper on the CPU + GPU than arbitrary re-sampling (which it used to do).  To compensate for each 1/2 zoom level the waterfall is actually twice the resolution internally vs. what's displayed and when you cross that resolution boundary that's when the decimation kicks in another level.

src/CubicSDRDefs.h
```diff
@@ -31,15 +31,30 @@ const char filePathSeparator =
...
+#define DEFAULT_SCOPE_FFT_SIZE (DEFAULT_FFT_SIZE / 2)
+
+//Both must be a power of 2 to prevent terrible OpenGL performance.
+#define DEFAULT_MAIN_WATERFALL_LINES_NB 1024
```
cjcliffe: Going to knock this back down to 512; I like that it shows more but it's really flickery when you reduce the height a bit.  I think we can make this configurable and even rig a 1:1 pixel match with multi-buffer scroll-back eventually.

src/CubicSDRDefs.h
```diff
@@ -31,15 +31,30 @@ const char filePathSeparator =
...
+#define DEFAULT_DEMOD_WATERFALL_LINES_NB 256
```
cjcliffe: This one looks better like this; it was always a bit stretched

cjcliffe:
All looks good; merged.

src/process/FFTDataDistributor.cpp
vsonnier: Thanks for the explanation :+1:

PR #515, opened 2017-02-15
Replace ThreadQueue by a full blocking queue.

vsonnier:
@cjcliffe Helllo ! This one is an attempt to turn the ThreadQueue into a fully blocking queue, named ThreadBlockingQueue. It is more complete, because the producer ``push()`` calls are now blocking by default when the queue is full, just like ``pop()`` is when the queue is empty.

Indeed, until now ``ThreadQueue.push()`` wasn't blocking on a full queue, forcing us to either explicitly stop pushing elements or simply discarding them on the producer side of the producer-consumer duo.

With a blocking ``push()`` by default, the threads communicating by a cascading ``push()/pop()`` are able to self-regulate their processing flow, while simplifying the code because nothing is lost in case of blocking ``push()`` by definition.

As a first attempt, I then replaced every ``ThreadQueue``  by  ``ThreadBlockingQueue``, so turning auto-magically all non-blocking ````push()```` by blocking ones.
The important difference to note is now that ``set_max_num_items()`` is very important to set explicitly, because the default value is 1.
As for the behavior with such radical changes, I've got some freezes on exiting application, and waterfall/scope stop refreshing.
So I instrumented the code to turn ``push(item)`` calls into timed-out ones : 
``push(item, 50*1000, "specific error message")``
so that the contention points revealed themselves.

In the end after re-introducing non-blocking calls at the contention points, the amount of non-blocking calls in the whole application was limited to:
- ``SDRThread::readStream()`` : make sense, drop samples from device if we cannot keep up with the realtime processing. 
- ``DemodulatorThread.cpp(302):  localAudioVisOutputQueue->try_push(ati_vis)``
- ``FFTDataDistributor.cpp(114):  distribute(outp, NON_BLOCKING_TIMEOUT);``

Now about the result: unless I was extremely lucky, I didn't see any application freeze, crash, or anything else after several days of torture loading, unloading sessions files, or in general usage whatsoever.
So, as for me it is better than before where freezes were still present.

Note that with ``ThreadBlockingQueue.try_push()`` the new class can do what the previous ``ThreadQueue.push()`` did, so can replace it no matter how we decide to use it. 

So please give it a try and tell me what you think !

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -442,6 +442,7 @@ SET (cubicsdr_headers
     src/util/Gradient.h
     src/util/Timer.h
     src/util/ThreadQueue.h
+	src/util/ThreadBlockingQueue.h
     src/util/MouseTracker.h
     src/util/GLExt.h
     src/util/GLFont.h
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -441,7 +441,6 @@ SET (cubicsdr_headers
     src/audio/AudioThread.h
     src/util/Gradient.h
     src/util/Timer.h
-    src/util/ThreadQueue.h
 	src/util/ThreadBlockingQueue.h
     src/util/MouseTracker.h
     src/util/GLExt.h
--- a/src/CubicSDR.h
+++ b/src/CubicSDR.h
@@ -12,7 +12,7 @@
 #include "GLExt.h"
 #include "PrimaryGLContext.h"
 
-#include "ThreadQueue.h"
+#include "ThreadBlockingQueue.h"
 #include "SoapySDRThread.h"
 #include "SDREnumerator.h"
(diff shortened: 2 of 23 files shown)


cjcliffe:
Looks great; going to try on Linux and do some more torture tests with the HackRF here and I'll likely merge it in tonight.

vsonnier:
@cjcliffe  I've restored non-blocking in 	051ebec for ``DemodulatorThread::audioOutputQueue->try_push(ati)``, because I remarked it could hang while debugging: Stop on break point, explore, step..etc. then continue running. At that point the whole application often hangs in a ThreadBlocking-deadlock problem :)

cjcliffe:
@vsonnier performance is at least on-par and everything seems stable here; going to merge it in

PR #542, opened 2017-04-22
Fix portselectordialog being added to sources and headers while USE_HAMLIB disabled

Dantali0n:


Dantali0n:
Oeps yes that's definitely a mistake, I have corrected the placement.

cjcliffe:
Whoops, missed that one thanks; need to remember to try building with it off too :)

PR #545, opened 2017-05-21
TheadBlockingQueue implemented with circular buffer (zero runtime mem allocs)

vsonnier:


vsonnier:
Well, looks like it works.

cjcliffe:
Nice work! 😃

vsonnier:
@cjcliffe  Thanks ! 
Funny thing, I'm going to throw away this work in the [vso_ReBuffer_with_shared_ptr](https://github.com/cjcliffe/CubicSDR/tree/vso_ReBuffer_with_shared_ptr) branch, because it doesn't play nice with ``std::shared_ptr<>``.
I originally wanted to replace the ``std::deque`` usage because I feared that in the common scenario ``1 element => empty => 1 element => empty...`` the ``std::deque`` would constantly allocating and deallocating memory because the container was empty.
After roaming the Internet, I saw that both MSVC and GCC implementations always allocate a minimum batch of entries, keeping it even when the container is empty. So we are fine.

Now if we start playing with a ``ThreadBlockingQueue< std::shared_ptr<> >``, the container have to be well behaved w.r.t destructors, so I would have to call elements destructors explicitly belonging to the underlying ``m_circular_buffer`` exactly like in a traditional container, and so on...

Indeed, using the circular buffer impl of ``ThreadBlockingQueue`` in the [vso_ReBuffer_with_shared_ptr](https://github.com/cjcliffe/CubicSDR/tree/vso_ReBuffer_with_shared_ptr) branch results that memory leaks because the m_circular_buffer kept references alive. (exact same problem as in Java !)

At that point I didn't want to complicate further the circular buffer impl, all the more that it didn't bring any visible performance gain compared to the usage of a ``std::deque``. Well, turns out STL is quite good, what a surprise.

As for the  [vso_ReBuffer_with_shared_ptr](https://github.com/cjcliffe/CubicSDR/tree/vso_ReBuffer_with_shared_ptr) branch, it works well enough with a slightly less CPU and memory consumption that some other bugs are showing up again, like the never-ending story of the hang application when involving ``DemodulatorMgr::terminateAll()`` with lots of demodulators...
So I'm still working on it.

PR #547, opened 2017-05-25
Rebuffer automatic refcount management

vsonnier:
@cjcliffe  Hello !
Time to another major break in the code.
This PR is about using ``std::shared_ptr`` in ReBuffer, completly removing the need of manual ``setRefCount()`` and ``getRefCount()``, which is error prone, and sometimes impossible to set properly according to objects real lifetimes.
So we gain:
-  Simplified ReBuffer usage, 
-  Correct ref-counting in all situations (ex: no longer concurrent set/getRefCount update troubles),
-  Somewhat better memory management and CPU consumption,
- Resurrected the previous ``ThreadBlockingQueue`` using ``std::deque``, needed to play nice with ``std::shared_ptr``. 
Indeed while used as ``ThreadBlockingQueue< std::shared_ptr< T > >``, the container have to be well behaved w.r.t destructors, so I would have to call elements destructors explicitly belonging to the underlying ``m_circular_buffer`` exactly like in a traditional container, and so on...

I observed that using the circular buffer impl of ``ThreadBlockingQueue`` in the [vso_ReBuffer_with_shared_ptr](https://github.com/cjcliffe/CubicSDR/tree/vso_ReBuffer_with_shared_ptr) branch created memory leaks because the m_circular_buffer kept references alive. (exact same problem as in Java !)

At that point I didn't want to complicate further the circular buffer impl, all the more that it didn't bring any visible performance gain compared to the usage of a ``std::deque``. Well, turns out STL is quite good, what a surprise.
 
Also in this PR I think I crushed the application hung problem for good. See commit comments:)

### DIFF
--- a/src/AppFrame.cpp
+++ b/src/AppFrame.cpp
@@ -1612,6 +1612,9 @@ void AppFrame::OnIdle(wxIdleEvent& event) {
         updateDeviceParams();
     }
     
+    //try to garbage collect the retired demodulators.
+    wxGetApp().getDemodMgr().garbageCollect();
+
     DemodulatorInstance *demod = wxGetApp().getDemodMgr().getLastActiveDemodulator();
 
     if (demod && demod->isModemInitialized()) {
--- a/src/IOThread.cpp
+++ b/src/IOThread.cpp
@@ -127,7 +127,7 @@ bool IOThread::isTerminated(int waitMs) {
         }
     }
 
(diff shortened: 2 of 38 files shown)


vsonnier:
@cjcliffe Fine, the frozen application when loading session files in succession with numerous demods (involving ``DemodulatorMgr::terminateAll()``) seems gone. 
Fortunately, such bugs have the habit of showing up ONLY after release. So trigger the Murphy's law by doing just that :) 
I also expect some reports on template syntax under *Nix and OSX.

PR #555, opened 2017-06-02
Change boolean to bool

lospheris:
boolean needs changed to bool for garbageCollect in both DemodulatorMgr.cpp and DemodulatorMgr.h. It causes a compile error since boolean in not a recognized keyword for C++. Compiles fine after changing both files.

Sorry for the 2 pull requests, I caught the dumb.


vsonnier:
@lospheris Thanks, I've done too much Java over the years that's why.

PR #571, opened 2017-08-25
Add  Antenna support (RX only), plus improved Settings menu.

vsonnier:
Hello @cjcliffe, @romeojulietthotel, @guruofquality, @SDRplay and all.

This PR brings 2 enhancements:

-  Antenna support (RX) through SoapySDR `listAntennas(SOAPY_SDR_RX,...)` and `getAntenna(SOAPY_SDR_RX,...)` calls. 
Now multiple-antenna systems can choose their starting RX Antenna, and later switch their RX antenna dynamically. 

SDRPlay RSP2 have it using the latest module master (tested with) and as far as I can see LimeSDR should have too. (others through SoapyOsmo ?)

In practice, multi-RX systems now see the following change at CubicSDR device selection:
device_selection
Note the addition of the `Antenna` property.

An 'Antenna' additional menu entry shows up in the `Settings` menu, from which you may switch the antenna dynamically:
new_settings

On systems with only 1 RX antenna, both `Antenna` additional entries are not present, so Cubic is the same as before to prevent UI clutter. 

In on the above image, you can see the second enhancement:
-  'Single-click' Settings view showing all settings current values on the `Settings` menu itself, so we no longer need to navigate the sub-menus of every setting just to know its current value.

### DIFF
--- a/src/AppConfig.cpp
+++ b/src/AppConfig.cpp
@@ -39,6 +39,14 @@ long DeviceConfig::getSampleRate() {
     return sampleRate.load();
 }
 
+void DeviceConfig::setAntennaName(const std::string& name) {
+    antennaName = name;
+}
+
+const std::string& DeviceConfig::getAntennaName() {
+    return antennaName;
+}
+
 void DeviceConfig::setAGCMode(bool agcMode) {
     this->agcMode.store(agcMode);
 }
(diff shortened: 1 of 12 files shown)


cjcliffe:
@vsonnier looks great; thanks for the addition!

vsonnier:
Welcome back @cjcliffe :)

PR #572, opened 2017-08-27
Automatic DemodulatorInstance memory management through shared_ptr and misc.

vsonnier:
@cjcliffe  Hello! This PR brings DemodulatorInstance automatic memory management using ``std::shared_ptr<DemodulatorInstance>` encapsulation by CubicSDR code.
This PR brings the following benefits:
-  No longer need to `garbageCollect` manuallly or anything,
-  Always correct deallocation, authorized only when no code is using the instance,
-  SDRPostThread simplification: no need to maintain a private copy of demodulators that can turn de-synchronized in case of problems,
-  Thanks to `std::shared_ptr`, in the worst case `DemodulatorMgr::getDemodulators()` returns a not-up-to-date list of demodulators, but all of them are actually (valid/still allocated) demodulators for the client code,
-  Misc: liberal use of `auto` in the modified code. 

A very important exception to the use of `std_shared_ptr` is in `DemodulatorPreThread` and `DemodulatorThread` where DemodulatorInstance parent must stay a raw pointer, because of the circular dependency. This is perfectly fine, because those thread lifetimes are completely managed by the  DemodulatorInstance itself.

### DIFF
--- a/src/AppFrame.cpp
+++ b/src/AppFrame.cpp
@@ -1709,10 +1709,7 @@ void AppFrame::OnIdle(wxIdleEvent& event) {
         }
     }
     
-    //try to garbage collect the retired demodulators.
-    wxGetApp().getDemodMgr().garbageCollect();
-
-    DemodulatorInstance *demod = wxGetApp().getDemodMgr().getLastActiveDemodulator();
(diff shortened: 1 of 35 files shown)


cjcliffe:
@vsonnier nice work; thanks again!

Issue #583, opened 2017-10-22
v2.0 no "Audio Output" Menue on MacOS Sierra (10.12.6)


aneitzel:
Hi,

on latest CubicSDR version i dont have menue "audio output" to select the audio device. I tested it with older v1.20 - with this it works and audio menue is shown as expected.

I would pipe sound thruth "Loopback" device to a digital decoder (DSD+).
Seems to be a bug :-(

vsonnier:
The audio menu no longer exists, audio output has been moved : see in  http://cubicsdr.readthedocs.io/en/latest/application-window.html "Modem properties" where you'll find it again.

vsonnier:
@cjcliffe Hello ! A side note: maybe you should modify your site www.cubicsdr.com a bit to point on GirHub releases because plenty of people do not seem to go beyond the original site and so use an outdated version.

aneitzel:
Thanks for this hint.... i was not aware of this change. Its ok and working!

cjcliffe:
@vsonnier I'll see what I can do about making it publish the releases automatically; should probably get a new one up soon anyways with all the latest fixes applied

PR #594, opened 2017-12-31
Audio recording

cjcliffe:
Audio Recording features
- Recording Path setting in File menu
- WAV File output support
- Internal AudioSink demod support
- Record/Stop buttons added in bookmark view actives list
- Activate Hover + R to record demod, Shift-R to record/stop all


vsonnier:
Hello @cjcliffe! First, I wish you a happy new Year !

I tried the `audio_recording` branch, and it works well here on Windows 10 x64.

I have a couple of UI suggestions:
-  Make the recording path appear in the menu entry itself, just like I did for Sample Rate /Manual entry.
- I think the pulsating of the frequency label during the is too faint. Maybe the whole spectrum band should pulsate, a bit like the squelch break make the whole waterfall band pulsate in yellow ?  
-  with the optional [V], [M], [S] and now [R] options in Waterfall, the frequency display is quit crowed and so overlaps very often, especially in non-default text sizes. Maybe we should move them elsewhere, for instance at the top of the label area ?

src/demod/DemodulatorInstance.cpp
```diff
@@ -580,6 +619,71 @@ ModemSettings DemodulatorInstance::getLastModemSettings(std::string demodType) {
...
+
+    time_t t = std::time(nullptr);
+    tm ltm = *std::localtime(&t);
+    
+    std::stringstream fileName;
+    
+    std::wstring userLabel = getDemodulatorUserLabel();
+
+    // TODO: Can we support wstring filenames for user labels?
+    std::string userLabelStr(userLabel.begin(), userLabel.end());
```
vsonnier: A better and  still quick way to convert `std::wstring` to `std::string` cleanly is to use a `wxString` that way:
````cpp
wxString userLabelForFileName(userLabel);
std::string userLabelStr = userLabelForFileName.ToStdString();
````
It works well to create non-ASCII file names where the previous solution creates invalid filenames.

**Comment by vsonnier** [comment]
Some thoughts about my previous suggestions:
2. After practicing, I'm getting used to the frequency-only green pulsing label. Since I'm always using 1.5x text size I find it good enough. So it is your call to improve that.
3. Instead of moving the flags elsewhere, I think compacting their display a bit would do. For exemple, instead of having `[V] 100.40 [R]` with spaces between, we could change it be like `[VRMS]100.4` i.e always start with the flags, stick the one-letter flags together, and use no space between the flags and the frequency. 
To me it seems easier to read the whole thing with all the flags at the beginning, if any.

If you agree with this I can contribute the 1. 3. 4. modifications if you don't have time to do it yourself.

**Comment by cjcliffe** [comment]
@vsonnier Happy New Year! -- all good suggestions thanks; I'll tinker with the placement / recording indication ideas when I'm back later this evening.  At some point I'd like to use some symbols/graphics to indicate things and will likely add a simple 'sprite sheet' type rendering which would be basically the same as the text renderer without the string building logic and font loader.

**Comment by vsonnier** [comment]
Last changes Looks Cool To Me.

**Comment by vsonnier** [comment]
@cjcliffe  I've found a bug, or a limitation if you wish : nothing is registered in the Wav file either if the demod is Mute or the level is below the squelch level. 

If you test by Mute or changing the squelch during the recording, and back again, you can hear that no "silence" is registered between the 2 sequences. 

If it doesn't ever be un-Mute or never break the squelch during the recording, the file is never created at all.

File: src/demod/DemodulatorThread.cpp [hunk 2]
```diff
@@ -224,17 +224,35 @@ void DemodulatorThread::run() {
...
+
             for (auto data_i : ati->data) {
                 float p = fabs(data_i);
                 if (p > ati->peak) {
                     ati->peak = p;
                 }
             }
         } else if (ati) {
+			//squelch situation, but recording is on-going, so record "silence" to AudioSink: 
+			if (localAudioSinkOutputQueue != nullptr) {
```
vsonnier: I case of squelch situation, I patch the `ati` samples to zero, keeping the size so the WAV timecode is respected.

src/demod/DemodulatorThread.cpp
```diff
@@ -334,7 +345,7 @@ void DemodulatorThread::run() {
             }
             
             if (localAudioSinkOutputQueue != nullptr) {
-                if (!audioSinkOutputQueue->try_push(ati)) {
+                if (!localAudioSinkOutputQueue->try_push(ati)) {
```
vsonnier: Small fix here, use the local variable. Besides, the code is made such that audio is recorded irrespective of the muted state of the demod, is this intended ?

starvald:
@vsonnier is recording hours of silence the intended result of these patches or are you still working towards something?

vsonnier:
Well, recording silence as you say is perfectly intended indeed. 
Imagine you start a recording, take note of the time `Start_time`, then stops 10 hours later (we now have terabyte HDs do we ?) 
Later on Iif you have something interesting at some date D in the recording, you will be able to know when it actually happened, i.e at `Start_time + D`. 
If you don't record "the silence", the the time will be all fucked up.

I checked that a 48KHz - Stereo (for FM) WAV file is roughly 0.2MB / s which leads to 720MB per hour. 
All the other modulations (NFM, AM and so on) are twice smaller because they are inherently Mono. 
I think that manageable.

vsonnier:
@starvald  To answer your question, I don't think there is more we can do here with a WAV file format so I'm not planing any further improvement on this.

mmgarland3:
Something I have been exploring lately is the notion of embedding tomecodes
into the recorded stream for forensic integrity.  I was looking at the FLAC
seektable and application metadata frames for this, specifically to record
RF streams and trim out the "silence".  Keep in mind, as a user, that any
recording method that excludes "silence" means squelch adjustment or you
will inadvertently cut off audio you intend to keep or include audio you
intended to discard.  It is better from an acquisition standpoint to record
an uninterrupted stream and then let the user trim out undesired data from
a copy of the file, and I believe that's the best option for CubicSDR as
well, though an option that only records when squelch is broken and writes
a new file to disk when squelched for more than, say 5 seconds, could be
useful for some applications.

One other caveat with .wav is that it is 32-bit constrained and files
cannot exceed 4GB, roughly 6 hours of 44khz stereo audio.

starvald:
couldn't you make that a toggle to record silence or not?
i'm recording 9 NFM channels 24/7 on laptop with an ssd and cubic is capable of being run on a Pi3 off a microsd card.
flash media will take a big hit with this.

vsonnier:
Thanks @mmgarland3 for the suggestions ! I merely fixed a inconsistency to be able to track the time correctly.
The user can define a squelch per-modem (or not), so the present way of recording excludes what the user didn't wanted to hear the same way. 
On the other hand, a muted demod is recorded anyway with the actual code.

It is up to @cjcliffe now to decide of further improvements among all those suggestions.

vsonnier:
@starvald The WAV is a simple "stream" format with the samples stored in a single array without other additional information. As a consequence, we are forced to store "zero samples" to mark the correct elapsing of time.

vsonnier:
In the last commit, I've removed the recording duration limitation by generating multiple files automatically:
- Limit WAV size to 2GB for maximum compatibility,
- Continue recording on another file when size gets too big (XXX_001.wav, then XXX_002.wav and so on)
- The sequence assure up to 2000GB worth of recording which should be enough for anybody.

That won't make WAV files any smaller, though :)

cjcliffe:
@vsonnier thanks for the patches; I hadn't tested large files before and forgot about the size limit; and after some consideration I also agree that the squelch shouldn't affect recording times by default..  

Though since @starvald 's use case matches my own initial thinking when deciding to discard the squelch gaps I think I'll go ahead and move the Recording path option into a "Recording" sub-menu and add a simple "Squelch" -> "Record Silence / Skip Silence / Record Always" option toggle there as well.

We do need to consider sometime adding a real "preferences" dialog to contain optional settings to soften some of the hard-coded details in general (and before we fill up the menu bar).

vsonnier:
@cjcliffe My thoughts exactly, about a recording option, and about confuguring squelch as well so tthat it is ok for all tastes.

vsonnier:
@starvald Hello ! I introduced the following changes in the last commit:
- Moved recording options to a dedicated menu,
- Added the Squelch options:  Record Silence / Skip Silence / Record Always. Happy now ?
- Added a file time limit, effectively closing #583. Now surely happy, you are.

starvald:
Sounds great, thanks @vsonnier :)
although it's failing to compile atm.

`[lee@Plasma build]$ make
[  1%] Building CXX object CMakeFiles/CubicSDR.dir/external/rtaudio/RtAudio.cpp.o
[  2%] Building CXX object CMakeFiles/CubicSDR.dir/src/CubicSDR.cpp.o
In file included from /run/media/storageEXT4/arch-compiles/CubicSDR/src/modules/modem/Modem.h:8:0,
                 from /run/media/storageEXT4/arch-compiles/CubicSDR/src/modules/modem/ModemDigital.h:5,
                 from /run/media/storageEXT4/arch-compiles/CubicSDR/src/demod/DemodulatorInstance.h:11,
                 from /run/media/storageEXT4/arch-compiles/CubicSDR/src/demod/DemodulatorMgr.h:10,
                 from /run/media/storageEXT4/arch-compiles/CubicSDR/src/visual/PrimaryGLContext.h:14,
                 from /run/media/storageEXT4/arch-compiles/CubicSDR/src/CubicSDR.h:13,
                 from /run/media/storageEXT4/arch-compiles/CubicSDR/src/CubicSDR.cpp:17:
/run/media/storageEXT4/arch-compiles/CubicSDR/src/audio/AudioThread.h:24:2: error: ‘boolean’ does not name a type; did you mean ‘GLboolean’?
  boolean is_squelch_active = false;
  ^~~~~~~
  GLboolean
make[2]: *** [CMakeFiles/CubicSDR.dir/build.make:87: CMakeFiles/CubicSDR.dir/src/CubicSDR.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:100: CMakeFiles/CubicSDR.dir/all] Error 2
make: *** [Makefile:130: all] Error 2
`

vsonnier:
LOL my Java habit strike again ! here it is.

vsonnier:
@mmgarland3:
> ...though an option that only records when squelch is broken and writes
a new file to disk when squelched for more than, say 5 seconds, could be
useful for some applications.

Now that the record file time limit is configurable, we have this kind of feature. 
The recording is implemented in such a way that no file is actually created, and named with the current timestamp until there is actually something to record.
For instance, say that you configure your File Time Limit to 10 s with the Skip Silence option on and start a recording for the night.

The next day, you will only get files whose names denote the actual date (in their timestamp) at which something break the squelch with a margin of  error of 10 s. 

From within a unique 10 s file, you won't know the exact time a given event occurred because of the reasons given above, but you won't record _hours of silence_ either. ((c)@starvald). 
That will indeed conserve a lot of disk space if the squelch breaks are rare.

vsonnier:
As you see I didn't use a dialog for Recording configuration because I think it is enough to have a separate menu considering we have very few options. Maybe in the future a Dialog may be needed if tons of other options are added.

cjcliffe:
@vsonnier thanks for the patches; taking a look now and will see about getting a build ready

cjcliffe:
@vsonnier seems to have introduced a crash on exit in OSX after any recording operations, attempting to track it down now:
`Terminating Visual Processor threads..
Application termination complete.
Assertion failed: (e == 0), function ~recursive_mutex, file /BuildRoot/Library/Caches/com.apple.xbs/Sources/libcxx/libcxx-120.1/src/mutex.cpp, line 82.`

cjcliffe:
@vsonnier looks like there's some sort of recursive mutex issue with pushing the same object to the sink thread unless it's at the end of the demod thread function -- if it's pushed before the viz section  uses it then it crashes with that error on application exit.

parkerlreed:
Just wanted to say, thanks a ton for the recording! Never realized how useful it would be and I love the implementation.

Issue #716, opened 2019-03-09
Sampling problems with LimeSDR at sample rates higher than 3Mhz


Dantali0n:
Whenever the sampling rate is increased above around 3Mhz it fails to sample the data. The output can only be recovered by stopping the device and restarting at lower sampling rate. My machine should be capable of handling these data rates having a 7700hq and a 1050 TI.

The outputs shows the following endlessly in a loop:
```
SDRThread::readStream(): 2. SoapySDR read failed with code: -1
SDRThread::readStream(): 2. SoapySDR read failed with code: -1
SDRThread::readStream(): 3.1 iqDataOutQueue output queue is full, discard processing of the batch...
SDRThread::readStream(): 2. SoapySDR read failed with code: -1
```

SoapySDR version: 0.7.1
LMS7Soapy version: 18.10.1
CubicSDR version: c561292

vsonnier:
Hello @Dantali0n, was the problem present if you revert CubicSDR for instance to 792bf20a9c42e634031f4ef2ad6812f17cdf7b69 (a.k.a before my Spin-lock experiments)

Otherwise the `-1` return code means `SOAPY_SDR_TIMEOUT`  and is returned by the Soapy module itself, and the problem is likely to come from there.

Dantali0n:
> Hello @Dantali0n, was the problem present if you revert CubicSDR for instance to [792bf20](https://github.com/cjcliffe/CubicSDR/commit/792bf20a9c42e634031f4ef2ad6812f17cdf7b69) (a.k.a before my Spin-lock experiments)
> 
> Otherwise the `-1` return code means `SOAPY_SDR_TIMEOUT` and is returned by the Soapy module itself, and the problem is likely to come from there.

No, I was waiting to see if the problem would persist after your spin-lock refactoring. 

Regarding the Soapy module I can use almost any Soapy interfacing software such as sdrangel but also the python modules of Soapy at the full 65Mhz rate perfectly.

vsonnier:
Right, I've added traces in `master` about the used MTU size usage. Can you test and report the ones you are getting depending of the sample rate ?
Maybe the problem is a too-small reported `getStreamMTU`, resulting too much `readStream` on the Cubic side, since it uses MTU to know how many samples are fetched at a time.

Other clients may not obey the MTU size ans request bigger chunks, resulting less overhead or problems.

Dantali0n:
Anything between 100Khz and 3Mhz has the same MTU size of 4080
[...]

PR #717, opened 2019-03-09
Fix for LimeSDR-USB at sample rates above 3Mhz #716

Dantali0n:
This is a fix to allow CubicSDR to use the LimeSDR-USB at sampling rate above 3Mhz.


vsonnier:
Your branch were pulling very old commits, so I simply cherry-picked your change on the `master`. 
We should not do that kind of device-specific patch often though, although there was already one for `BladeRF` located at the same place so that's not so ugly.

Thanks @Dantali0n for your contribution !
 
Closing this pull request.

Issue #703, opened 2019-01-18
cubicsdr0.2.5 on Debian Stretch - Unable to enter serial port  


ajcher:
Built CubicSDR on Debian Stretch and got it working. Next step is to get cubicsdr to talk with rigctld so cubicsdr can follow the rig. Unable to change (i.e. manually enter) serial port info (refer image). 

In CubicSDR, RigControl->Enable Rig is selected and the RigControl->Model is set to Hamlib Net Rig Ctl. Need to enter 127.0.0.1:nnnn. Is there a way to do it in the config file?

![image](https://user-images.githubusercontent.com/46802452/51387798-68425c80-1af5-11e9-940c-b51920c95407.png)

Thx a lot.

cjcliffe:
@ajcher there should be a text box just below the list where you can type manually; does it appear if you resize the window?

cjcliffe:
@ajcher ah, yes that should be resizable; and probably larger by default.  Thanks for the follow-up.

ajcher:
Found a workaround ... by  forcing the ip:port sequence in the config file...but the config file gets over-written each time i use the pgm. The way out was to make the necessary changes in settings, quit pgm, hard code the ip:port sequence and change the file permission to RO. Which means, it kinda-sorta works and am functional for now. Thx for all the effort you put in to keep this going. Much appreciated ...

cjcliffe:
@ajcher latest merge to master branch now includes a fixed port selection dialog with resize supported; you should now be able to use the text box below the entry list to set the ip:port

Issue #725, opened 2019-05-04
Unable to enter serial port and compilation without Rigcontrol


magellan-13016:
Hello,

I'm using Linux Mint 18.3 and I've the same problem with AppImage as described before : There is no way to enter serial port and I need it for Doppler Correction with GPredict. 

So I've tried to compile the software by following this link : 
https://github.com/cjcliffe/CubicSDR/wiki/Build-Linux

After a long time, I finally succeed but I was disappointed because there is no "RigControl" in the menu.

My goal is to use CubicSDR to make a IF Panadaptor for my Yaesu FT-847 but before buying the card to insert in my transceiver. 

Thanks for help.

David.

cjcliffe:
@magellan-13016 for your CAT control issues you mentioned in PR #728 that you can enter the port but can't connect; can you confirm that your user has serial port device access?  I found in Ubuntu here without setting up user device rules I had to run CubicSDR as root (via sudo) to access the serial ports.  

You should also confirm that the radio is actually communicating over serial and the appropriate cable is being used (my Uniden scanner needed a 'NULL MODEM' type serial cable and specific baud rate).  If you have another program that's working with your radio's CAT to confirm functionality it would be helpful.

Otherwise you should be able to see messages in the console (if you launch CubicSDR from the command line) which would likely display an error message from hamlib about permissions, inability to connect or other errors when initializing the rig.

As far as hooking in GPredict I'm not sure what to do; you should be able to get CubicSDR to track a remote program via the rig control; I've done it using FLDigi and the 'Hamlib NET rigctl' before where I was able to have CubicSDR follow the frequency control from FLDigi  --  see https://github.com/Hamlib/Hamlib/wiki/FAQ#i-have-2-programs-which-use-hamlib-can-i-use-them-with-the-same-radio and https://github.com/Hamlib/Hamlib/wiki/Network-Device-Control for some basics on using rigctl.

Also I checked the Wiki and the line "Note: add -DUSE_HAMLIB=1 to cmake command line to include hamlib support." is already written directly below the "Building CubicSDR" header -- if there's something I can do to make it more visible then suggestions are welcome :+1:

cjcliffe:
@magellan-13016 here's a fresh AppImage build for testing:

[CubicSDR-0.2.5-x86_64_AppImage.zip](https://github.com/cjcliffe/CubicSDR/files/3160215/CubicSDR-0.2.5-x86_64_AppImage.zip)

Cheers.

[...]

PR #728, opened 2019-05-08
Fix port selector dialog; simplify rig model selection with manufacturer sub-menus

cjcliffe:
- Rig serial port selector dialog layout errors fixed (#703)
- Dialog can now be resized
- Rig model menu will group models for each manufacturer into sub-menus if more than one model per manufacturer


cjcliffe:
@magellan-13016 Glad to hear port selector confirmed working -- still a short list of issues remain until the 0.2.6 release but in the meantime I'll try spinning up a new AppImage and post it in your #725 issue.

PR #748, opened 2019-08-24
Use Google Turbo Color for Default Theme WF colors

vsonnier:
Hello @cjcliffe,

This is a simple change for the Waterfall colors of the Default scheme using the Google Turbo Color scheme : https://ai.googleblog.com/2019/08/turbo-improved-rainbow-colormap-for.html (Apache 2.0 license)
 
I think it looks  good, but I'll let you decide on this one.

Regards,

Vincent


vsonnier:
@cjcliffe Respectfuly ping on this one. Google Turbo colors or not ?

vsonnier:
Attempting screenshots...
Normal Default Theme:
Before_WF

Default Theme with Turbo colors:
After_WF

Dantali0n:
I like the scheme but personally I think it should be up to the user to decide. Although the turbo schema overs smoother transitioning between colors which  makes it looks more modern and elegant it loses some contrast around specific areas.

Typically, I set up the waterfall precisely so back ground noise just stays in the blue region. Any signal coming up above the noise floor will immediately transition into green due to the sharp contrast around that region in the current color map.

However, When observing the two examples provided above to me the third signal of the left is clearer in the turbo color map than with the jet colormap. So it depends on how you configure it I suppose. While CubicSDR offers relatively little control over the signal levels required to produce specific parts of the colormap.

vsonnier:
Hi @Dantali0n !

> I like the scheme but personally I think it should be up to the user to decide.

Add another theme is easy, "Default with turbo colors" but will it be used ? I have no idea of the proportion of users even aware that they can change graphics like colors themes, not speaking of font size.  Maybe I'm biased because I'm only judging by what I see on social networks and Youtube, I don't know.

> Typically, I set up the waterfall precisely so back ground noise just stays in the blue region. Any signal coming up above the noise floor will immediately transition into green due to the sharp contrast around that region in the current color map.

Good reason indeed, but put it the other way: turbo colors on the other hand make it easy to discover weak signals close to more powerfull ones that otherwise would be less visible with Cubic "dynamic" contrast. You say that much on your third paragraph ir seems.

A compromise would be:
- Change Default Theme  with Turbo ,
- Add another theme "Default (Jet waterfall)"

An advantage would be that Cubic users would see the difference with previous versions, and at the same time discover the existence of Color Themes. Maybe.

The whole idea came to me only because I've seen some Tweet announcing Turbo Colors would become default colors on SDRConsole, so I discovered this color scheme and tested it.

Dantali0n:
> A compromise would be:
>     * Change Default Theme  with Turbo ,
>     * Add another theme "Default (Jet waterfall)"

I love the idea it will show users the new theme and if they really dislike it I am sure they will search for a way to change it back and either find it in the settings by themselves or they will open an issue here. Either way we can easily tell the user how to set it back and they will automatically discover the new color scheme when updating.

vsonnier:
Added `Default (Jet Waterfall)` as the "old" default theme.

vsonnier:
OK, let's push this :)

PR #767, opened 2019-11-03
Fix missing import for GLX on linux.

benallard:
Without this Display (for instance) is not defined in the scope.


vsonnier:
Thanks @benallard for your contribution !

PR #772, opened 2019-12-09
Winsock.h was included multiple times through windows.h causing compile errors

antihax:
Added `#define _WINSOCKAPI_` to prevent windows.h importing winsock.h and removed a few redundant windows.h imports.


vsonnier:
Hi @antihax , Visual Studio 2019 works for me out of the box, like 2017 and 2015 did worked out of the box before. What configuration are you using by any chance ?

If it is something esoteric like Mingw32 or MSYS2 I'm no even sure it is supported anymore.

Anyway, some remarks about the PR:
- We are not going to patch external OSS, even embedded like `RtAudio` is.
- Sound strange to put WinSocket specific hack in source code that don't use it at all. Again something may be wrong with your configuration.

antihax:
Interesting, I am using VS2019. Are you defining `WIN32_LEAN_AND_MEAN` anywhere?
I will take another look tonight where this is coming from.

vsonnier:
> Are you defining WIN32_LEAN_AND_MEAN anywhere?

I'm not playing with low-level settings like that, because it is irrelevant with CMake. I just generate the adequate VS solution with CMake following the steps and settings of the Wiki. 
Please note that this is the only supported way: VS solution generated with CMake as Wiki describes, with wxWidgets 3.1.3 statically build, while using the the config `Release` (or `RelWithDebInfo` if you want to debug CubicSDR)...etc.

For any other config variant you are on your own, and we are not going to dedicate any time to support such variants, sorry.

antihax:
Unfortunately I also did the same, but CMake output will not build due to redefining winsock.h imports.

I will see if I can identify if I did something incorrect when generating the project from CMake.

vsonnier:
I use a recent CMake, like 3.15+. (3.16 at that moment). Try to delete the CMake cache (see CMake GUI) and start again from the begining.

antihax:
I checked I have the latest versions of all components, latest Win10 SDK, and completely reset all MSVC 2019 settings. Problem persists on all imports of windows.h.

I create a quick test solution with nothing else and can successfully import windows.h with no issue. Not sure what in CubicSDR or a dependency is creating the conflict, but something is importing winsock out of order.

I've refactored to include WIN32_LEAN_AND_MEAN in the pre-processor options, which is a cleaner solution.

vsonnier:
> I've refactored to include WIN32_LEAN_AND_MEAN in the pre-processor options, which is a cleaner solution.

Great, works for me too ! Please re-add `Timer.cpp` Win32 include removal, and we are good to go.

vsonnier:
@antihax Thanks for your contribution(s) !

PR #773, opened 2019-12-09
Stop spamming hundreds of `No SoapySDR Module` dialogs.

antihax:
Not entirely sure how this was occurring as apparently you can't race inside wxTimers, but setting `failed` before creating the dialog stopped this bug.

Setting it after caused the dialog to be created hundreds of times.

Make sense if `ShowModal()` is a blocking call.


vsonnier:
@antihax Thanks !

PR #806, opened 2020-04-06
Adds a fix to scale the Frequency dialog.

jawatson:
The dialog layout uses a sizer to scale the content automatically.  The layout wasn't very good using sizers alone as width of the dialog was being set too narrow for the title string.  In the event that the initString is too short, the width is set manually to either the [title string width + 35%] or the [content string width +10%], whichever's the greater.  If the initString is very long, the sizer defines the width of the box.

This has been tested on Linux (Fedora 32) in both conventional and HiDPI (scale 200%) environments.  

This has not been tested in either Windows or Mac environments.

### DIFF
--- a/src/AppFrame.cpp
+++ b/src/AppFrame.cpp
@@ -1404,7 +1404,13 @@ bool AppFrame::actionOnMenuReset(wxCommandEvent& event) {
         wxGetApp().getSpectrumProcessor()->setFFTAverageRate(0.65f);
         spectrumAvgMeter->setLevel(0.65f);
 
-        SetTitle(CUBICSDR_TITLE);
+        wxString titleBar = CUBICSDR_TITLE;
+        //append the name of the current Device, if any.
+        if (wxGetApp().getDevice()) {
+            titleBar += " - " + wxGetApp().getDevice()->getName();
+        }
+
+        SetTitle(titleBar);
         currentSessionFile = "";
 		currentBookmarkFile = "";
         bookmarkSplitter->Unsplit(bookmarkView);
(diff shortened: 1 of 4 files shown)


vsonnier:
Hi @jawatson thank you for the fix, unfortunatly it has several problems:

- The dialog is no longer centered because `Centre()` was removed
- It doesn't work in Windows 10 , the `GetTextExtent` apparently measure up to the dot '.' of the number, wrongly computing the length.
- As far as I see (I'm not proficient in wxWidgets at all) we may be able resize this dialog, but the control do not appear at all ? again on Windows 10. 

Improvement suggestions:

- The Edit label dialog (E) should be plagued by the same problems as the Frequency Dialog, with hardcoded dimensions and should be modified the same way the Frequency dialog was.
 
Sorry to ask you all this, while I have very little time to investigate myself !

Dantali0n:
I will try this patch as well (Linux, i3) but can't test the scaling as my window manager simply does not support it (xrandr does though).

Dantali0n:
Screenshot from 2020-04-12 17-47-43

Looks fine on i3 at 1920x1080 (476mm x 268mm).

vsonnier:
Hi @jawatson, I pushed on master a fix that seems to work for both Frequency and Edit Label dialogs, based  on your code. Thanks for your contribution !

Dantali0n:
> Hi @jawatson, I pushed on master a fix that seems to work for both Frequency and Edit Label dialogs, based on your code. Thanks for your contribution !

Did you know that if you add a Co-authored-by line to your commit message that Github will detect it and attribute as well. Here is an example: https://github.com/openstack/watcher/commit/aa2a084d220614e0db79b15e669fae2c2b3423b7

vsonnier:
> Did you know that if you add a Co-authored-by line to your commit message that Github will detect it and attribute as well. Here is an example: openstack/watcher@aa2a084

Thanks, however it does not seem to work...

jawatson:
Thanks for letting me know - I'm glad that the code was of some use.  This is my first time doing anything with wxWidgets but I'm keen to learn more.

PR #865, opened 2020-12-25
Modified ModemAM

colbyAtCRI:
Implement an AM demodulator in ModemAM module. Replaces ampmodem based code which is too narrowband to be usable in many situations. DSB demodulator is not modified and is essentially identical to the AM one replaced by this.


vsonnier:
Thanks @colbyAtCRI for your contribution, I'll look into it.

PR #867, opened 2021-01-03
Updating Hamlib to v4+

vsonnier:
Hamlib 4+ has been relreased : https://github.com/Hamlib/Hamlib/releases/tag/4.1

This is the update of Hamlib to v4+, in particular Windows binaries. 

So far I modified `CMakeLists.txt` to accomodate the lib name change (`-4` instead of `-2`) and added the Windows binaries, and one constant in a Cubic .cpp

So far it compiles (with warnings) both for Win32 and Win64 using VS2019 v16.5

Edit :
- 4.0 :  runs without apparent issues, but what should I know ? I don't have Rigs to test with, so I opened this PR :)
- 4.1 : print `2021-03-03:18:22:20.712339: rig_mW2power called` then segfault. :(
- 4.2 : same as 4.0 :)

I'm using in both cases the Win Dlls from the Hamlib releases .

Vincent


vsonnier:
Updated for v4.1, changed title and such.

vsonnier:
Hamlib v4.2 has just been released: it runs under Win32 / Win64 without crashing this time.

src/rig/RigThread.cpp
```diff
@@ -114,7 +114,7 @@ void RigThread::run() {
     std::cout << "Rig thread starting." << std::endl;
 
     rig = rig_init(rigModel);
-	strncpy(rig->state.rigport.pathname, rigFile.c_str(), FILPATHLEN - 1);
+	strncpy(rig->state.rigport.pathname, rigFile.c_str(), HAMLIB_FILPATHLEN - 1);
```
vsonnier, line 117: The only change I made on CubicSDR source is here:

Issue #341, opened 2016-05-04
Feature Request: CW modem / filter


neilwillgettoit:
Hi,

It would be very nice to have a cw modem. Right now the usb/lsb can be used to listen, but when incorporating rig control / hamlib, it ends up requiring a tuning offset to shift the carrier into an audio tone. To make the modem complete, it would be nice to have a user set side tone, such as 650hz 
- [x] CW Filtered Mode

TonMachielsen:
@cjcliffe Looks like the Digital Lab needs to get a kick in the butt again. Or just leave it out and rely on third party software using audio pipes.

vsboost:
I'm in no rush.

Toontje:
@cjcliffe Looks like the Digital Lab needs to get a kick in the butt again. Or just leave it out and rely on third party software using audio pipes.

neilwillgettoit:
@Toontje I'm not looking for any type of decoder. Just a button for cw on the mode list so I can tune in morse signals with a bit more resolution than a usb/lsb filter, and with the side tone offset so the frequency resolution matches my rig vfo.

neilwillgettoit:
For example: 

When listening to a busy band like 40m, you will get multiple stations that fit within the standard 2.7khz ssb filter: 

![screen shot 2016-05-04 at 8 30 15 am](https://cloud.githubusercontent.com/assets/17776082/15015667/de74121a-11d2-11e6-8ee4-ba4b398b0c4e.png)

Even if I chop the filter down, because it is only reducing the top end of the ssb window, I cannot filter out other stations that are near and below the signal I am trying to focus on: 

![screen shot 2016-05-04 at 8 31 40 am](https://cloud.githubusercontent.com/assets/17776082/15015700/0460d300-11d3-11e6-9a5f-28d8bf2d25ca.png)

And then there is still the issue with sidetone and how that conflicts with rig control.

cjcliffe:
@neilwillgettoit this is definitely something I'd like to try to do as a decoder eventually; @Toontje is on track as I've already kinda tried it with the liquid OOK decoder :)

I'll start with the CW filter mode and see what I can do about feeding it to the OOK decoder for interpretation in the future.

Toontje:
Just getting into CW now and i see what @neilwillgettoit means. I +1 the request for a CW modem. Currently what i do is set the filter width to 500 and listen in USB/LSB, but it's not ideal.

Toontje:
Bump?

[...]

PR #871, opened 2021-01-15
Add ModemCW #341

colbyAtCRI:
This adds ModemCW feature (see #341)

- user settable offset frequency (default 650Hz) range [200:1000]
- default IQ bandwidth 500Hz
- Auto audio gain user selectable (default "on")
- Manual audio gain setting (default 15 dB) range [0:40]


cjcliffe:
@colbyAtCRI this looks great; thank you -- just trying to fix the build system and then I'll get this tested and merged in

PR #885, opened 2021-04-05
Cleanup: C++11 updates, clean-up, warning fixes

cjcliffe:


Dantali0n:
These look good, personally I like the explicit constructors and adding the intent to function arguments although I am not sure they are needed in clear get / set functions. However, you could of course make the argument that consistency is more important.

Personally I'd wish C++ had function argument intent keywords with well defined restrictions and the subsequent compiler warnings / errors.

PR #892, opened 2021-05-20
modem volume controls

colbyAtCRI:
I've added a class ModemAnalogVC which has ModemAnalog as a base class. This adds user settable modem gains to any modem that uses this base class. I've applied this to AM, CW, DSB, USB, and LSB modems. Current default is to have a modem gain of 15 dB with Auto Gain off. 
 
Screen Shot 2021-05-20 at 4 34 38 PM


colbyAtCRI:
This is in response to#826

vsonnier:
Hi @colbyAtCRI could you please rebase your work on the current Cubic `master`, it will be easier to merge.

src/AppFrame.cpp
```diff
@@ -60,7 +60,7 @@ class ActionDialogBookmarkReset : public ActionDialog {
     ActionDialogBookmarkReset() : ActionDialog(wxGetApp().getAppFrame(), wxID_ANY, wxT("Reset Bookmarks?")) {
         m_questionText->SetLabelText(wxT("Resetting bookmarks will erase all current bookmarks; are you sure?"));
     }
-
```
Dantali0n, line 63: Could you see if you can remove all these stylistic / linting changes as they aren't really part of the issue as well as making it much harder ro review.

src/AppFrame.cpp
```diff
@@ -3122,8 +3127,8 @@ void AppFrame::toggleAllActiveDemodRecording() {
     }
 
     for (auto i : allDemods) {
-      
-        i->setRecording(stateToSet);
```
Dantali0n, line 3126: If you really want to improve tabbing, indenting or trailing spaces I'd be al for it but I think it should be in a separate merge request.

src/modules/modem/Modem.h
```diff
@@ -130,37 +130,40 @@ class Modem : public ModemBase  {
...
+
     Modem();
     virtual ~Modem();
-    
+
     virtual ModemArgInfoList getSettings();
     virtual int getDefaultSampleRate();
     virtual void writeSetting(std::string setting, std::string value);
-    virtual void writeSettings(ModemSettings settings);
+            void writeSettings(ModemSettings settings);
```
Dantali0n, line 146: Why are modem writeSettings and readSettings no longer virtual?

src/modules/modem/analog/ModemCW.h
```diff
@@ -2,53 +2,48 @@
...
     ModemCW();
-
     ~ModemCW();
 
     std::string getName();
 
     static ModemBase *factory();
 
-    int checkSampleRate(long long srate, int arate);
+    int checkSampleRate (long long srate, int arate );
```
Dantali0n, line 25: Could you remove trailing space after `int arate`

Dantali0n (commented):
Heej, overall this looks good to me there are some things I noticed such as a lot of linting changes that are unnecessary to implement volume control. I also think you accidentally made read/writeSettings of Modem non virtual.

I don't know how strict everyone else wants to be about this but I personally don't like 'side effects' of patches. I think if you want to improve the linting it should be done in a separate merge request.

src/modules/modem/Modem.h
colbyAtCRI: I assume I left off a virtual keyword?

src/modules/modem/analog/ModemCW.h
colbyAtCRI: An interesting question. Which linter are "we" using? I assume I need to edit and commit this as a change to the pull request fork. This raises several issues. Cubic no longer compiles on my MacPro. Makes it hard for me to vet even a minor change as this.

vsonnier:
Some generic remarks:
- The default setting should be Auto gain = on, as before
- The save/load mechanism is missing for session xml for the additionnal options (?)

vsonnier:
To be clear @colbyAtCRI  we really appreciate your efforts, here ! Unfortunatly there are lots of uninteresting little details to fix before the commit could be accepted, so that the code stays clean and consitent for the future. 
I have little time to work on Cubic these days, so I'm just giving cheap advice, sorry.

colbyAtCRI:
Okay, I think I did that correctly

colbyAtCRI:
Okay, so I'm a little uncertain as to my next move. The save/load feature I didn't implement should be added by all means. I'm uncertain on how this is done. 

I will try to compile the latest version of Cubic. If it compiles on my MacPro, I'm open to figuring the save/load stuff out.

Dantali0n:
Hi @colbyAtCRI Thanks for the patch, I'll try to isolate and remove the linting changes on my part and be sure to tag you as author in the commits.

Don't worry about save/load if you don't have time to figure it out / get a development environment going right now. We could do that in a later patch as far as I am concerned.

PR #927, opened 2021-11-16
Update RtAudio to v5.20

vsonnier:
With PulseAudio fixes, among others. Works out of the box for Win32/64 of course. Still, despite WASAPI improvements no need to switch from good old DirectSound by default on Windows since it works very well as it is.


vsonnier:
Let's go !

PR #935, opened 2021-12-29
Fix typing errors

dforsi:
Most errors are in comments.
Only 3 are in messages visible to users:
* src/CubicSDR.h
* src/visual/SpectrumCanvas.cpp
* src/visual/WaterfallCanvas.cpp

Found with:
codespell --skip=external --ignore-words-list=axix,bord,parm,soley,ue --write-changes --interactive 2
and a manual edit.


vsonnier:
Better English doesn't hurt, thanks for your contribution !

PR #937, opened 2022-01-02
Tag and Release v0.2.7a

SDRplay:


vsonnier:
Hello Charles @cjcliffe, wherever you are, hope you fare well. 

To complement @SDRplay I think it is time we tag and release a new version namely `0.27a`. 

Hi @guruofquality ! I would also like to release a fresh Win64 install of Cubic in the simple form of a Zip, but I have neither the patience nor the ability to re-generate all the SoapySDR `modules` for all the devices out there, as Charles did previously.

So @guruofquality if you agree, I would like to borrow your prebuilt `SoapySDR.dll` and `modules` and all dependent libraries you already generated for your PothosSDR distribution, while only the CubicSDR executable would be mine. 

You will be credited accordingly of course :)

Let's give us 1 week to receive all the answers then I'll tag an publish an new Release at a bare minimum.

And.. Happy  New Year to all !

Edit : changed PR title

vsonnier:
Well, that was a looooonnng week. I finally pushed the `0.2.7` (without the 'a' to be consistent with other releases) 

Congrats to all contributors, an people who has provided support on the issues ! (@Dantali0n and many others)

No response from @cjcliffe or @guruofquality yet, so it will be a source/tag only release for now.

consultants007:
Where can I download Cubic SDR V0.2.7 for Mac. I need a DMG file.

PR #1022, opened 2024-01-30
Speedup compilation on all platforms

GermanAizek:
More information about package:
https://ccache.dev


cjcliffe:
thanks for the patch

PR #1055, opened 2026-08-21
Fix build errors against wxWidgets 3.3

aliaga17:
## Problem

CubicSDR fails to compile against wxWidgets 3.3 (current Homebrew release on macOS). Newer wxWidgets removes implicit conversions that older releases tolerated:

- `wxT()` literals no longer concatenate with `std::string` / `std::wstring` operands (`AppFrame.cpp`, `BookmarkView.cpp`)
- `std::wstring` no longer streams into `std::cout` implicitly (`SessionMgr.cpp`, `GLFont.cpp`)

## Fix

Wrap the affected concatenations in explicit `wxString` constructors and convert wide strings before streaming to stdout. No functional changes; labels and log output are identical.

## Testing

Builds and runs cleanly on macOS (arm64) against Homebrew wxWidgets 3.3; menu labels, bookmark dialogs, and session/font log lines render as before.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/AppFrame.cpp
+++ b/src/AppFrame.cpp
@@ -1184,7 +1184,7 @@ void AppFrame::handleUpdateDeviceParams() {
         sampleRateMenuItems[wxID_BANDWIDTH_MANUAL]->Enable(false);
     }
     else {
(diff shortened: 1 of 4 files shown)


cjcliffe:
Looks good and still compiles with older wxWidgets -- thanks

